PR #8, opened 2024-03-22
Add all local files to gitignore

jgeaso1266:
Add secret local properties to gitignore and remove them from being tracked.


Nath-Assa (approved):
Updated gitignore

local.properties and app/local.properties must not be checked into Version Control Systems, as they contain information specific to our local configuration.

secrets.properties was already in the gitignore. It is crucial to keep it there, as it contains sensitive information.

PR #55, opened 2024-04-08
Remove Bootcamp-specific Code and Update Package Name

Nath-Assa:
**Overview:**
This PR represents a major cleanup, removing all remnants of the initial Bootcamp project that are no longer relevant to our application's current direction.

**Key Changes:**

- Removed Outdated Test Files: All test files associated with the bootcamp project have been removed, as they no longer reflect our application's current or future state.

- UI Cleanup: Deleted CreateTodoScreen.kt and EditTodoScreen.kt, moving past the to-do list functionality that was part of the bootcamp's scope.

- Data Model and ViewModel Updates: Removed classes ToDo.kt, TodoList.kt, ToDoStatus.kt, ToDoViewModel.kt, and Location.kt that were central to the to-do list feature but are now obsolete.

- Package Name Change: Updated the package name from com.swent.bootcamp to com.github.se.eventradar, aligning with our GitHub repository and enhancing our consistency.

- Preservation of Potentially Useful Code: Commented out sections of code in various files that, while currently obsolete, hold potential for adaptation and integration into future functionalities.


app/src/main/java/com/github/se/eventradar/map/Map.kt
```diff
@@ -1,41 +1,26 @@
...
-import com.google.maps.android.compose.GoogleMap
-import com.google.maps.android.compose.MapProperties
-import com.google.maps.android.compose.MapUiSettings
-import com.google.maps.android.compose.Marker
-import com.google.maps.android.compose.rememberCameraPositionState
-import com.google.maps.android.compose.rememberMarkerState
 
 @Composable
 fun Map(viewModel: MapViewModel = viewModel(), navigationActions: NavigationActions) {
+  // The following block is temporarily commented out as part of the transition from managing ToDos
```
jgeaso1266, line 9: nit: you can simply delete this block.

app/src/main/java/com/github/se/eventradar/map/MapViewModel.kt
```diff
@@ -1,53 +1,9 @@
...
 import com.google.firebase.firestore.ktx.firestore
 import com.google.firebase.ktx.Firebase
-import kotlinx.coroutines.flow.MutableStateFlow
-import kotlinx.coroutines.flow.StateFlow
 
 class MapViewModel : ViewModel() {
-  private val _uiState = MutableStateFlow(MapUiState())
-  val uiState: StateFlow<MapUiState> = _uiState
-
   private val db = Firebase.firestore
```
jgeaso1266, line 8: nit: This line can be deleted with everything else.

app/src/main/java/com/github/se/eventradar/model/OverviewViewModel.kt
```diff
@@ -23,56 +18,21 @@ class OverviewViewModel(db: FirebaseFirestore = Firebase.firestore) : ViewModel(
...
-          Log.d("OverviewViewModel", "Error getting documents: ", exception)
-        }
   }
 
   fun onTaskClicked(id: String) {
-    val toDo = _uiState.value.toDoList.getAllTask.find { it.id == id }
+    // Temporarily commented out as we transition from handling to-dos to handling events.
+
+    /*
+    val to_Do = _uiState.value.toDoList.getAllTask.find { it.id == id }
```
jgeaso1266, line 27: nit: there the naming convention `to_Do` is not standard compared to what it was before (`toDo`).

jgeaso1266 (approved):
LGTM (had a few nits but that's it)

app/src/main/java/com/github/se/eventradar/map/MapViewModel.kt
Nath-Assa: Agreed

app/src/main/java/com/github/se/eventradar/map/Map.kt
Nath-Assa: Agreed

achillet3: I would agree with Jalen since we can still access it via version control if necessary

app/src/main/java/com/github/se/eventradar/model/OverviewViewModel.kt
achillet3: It is indeed a bit confusing. However, I would argue that in Map.kt the naming convention is still 'toDo' and has not been changed. So either way is fine with me but the names should be consistent between OverviewViewModel.kt and Map.kt. In this case, I would recommend leaving everything as toDo even if it might not be the cleanest. As said above, these blocks can also be deleted for now since we have them in a previous commit.

jgeaso1266 (approved):
LGTM

achillet3 (approved):
All good on my side as well!

PR #60, opened 2024-04-11
Update the web client ID to reference the new Firestore Database

jgeaso1266:
## Objective
Fix bug caused by using old web-client-id from the bootcamp to link to Firebase. Also add a default web client id.

## Key Changes
- [ ] replace the old web-client-id with the new database ID
- [ ] add a default web-client-id

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1,6 +1,7 @@
 <resources>
     <string name="app_name">My Application</string>
-    <string name="web_client_id">957415054330-j5fa1cl19qlthoshv7l2cccqv5v4icj4.apps.googleusercontent.com</string>
+    <string name="web_client_id">1043895872028–8m8f4kpcmqe60ofkef734vbgqph8g5ut.apps.googleusercontent.com</string>
+    <string name="default_web_client_id" translatable="false">webClientId.apps.googleusercontent.com</string>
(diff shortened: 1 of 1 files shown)


nikhen-s (approved):
This looks great to me. I had to change these files in my recent commit too so that google firebase authentication worked. Thank you for creating a separate pull request to solve this specific bug, tried it in my local machine and works great too :)

Viccha18 (approved):
Thank you !
I will just point out that :
in Nikhen's he uses "client id"
`<string name="client_id">1043895872028-8m8f4kpcmqe60ofkef734vbgqph8g5ut.apps.googleusercontent.com</string>`
and in yours "web_client_id"
`<string name="web_client_id">1043895872028–8m8f4kpcmqe60ofkef734vbgqph8g5ut.apps.googleusercontent.com</string>`
If that changes antyhing

nikhen-s:
Thanks for pointing it out Vicky, I have updated my most recent commit to the Google authentication feature branch, to use web_client_id like Jalen. Both work fine but for consistency, we can use Jalen's strings.xml file now.

jgeaso1266:
I will close this PR so that Nik can use his changes. Nik, can you make a separate PR with just the changes related to fixing this bug?

nikhen-s:
Yes will do so.

PR #65, opened 2024-04-11
Static Event details screen

Viccha18:
# Objective
Create a screen that shows more details about an event.

# Details
The screen is currently static with fixed values, its does not load any data from firebase
- [ ]     Create a new screen that holds the event details
- [ ]     Create a Register button that allows users to register for the event (button is here but triggers nothing)
- [ ]     Create a back button that takes users back to the previous screen in the navigation (implemented)


app/src/main/java/com/github/se/eventradar/ui/event/EventDetails.kt
```diff
@@ -30,210 +30,137 @@ import androidx.compose.ui.text.font.FontWeight
...
+
+    // Go back button
+    Button(
+        onClick = { navigationActions.goBack() },
+        modifier = Modifier.wrapContentSize().align(Alignment.Start).testTag("backButton"),
+        colors =
             ButtonDefaults.buttonColors(
-                contentColor = DT_titleColor,
+                contentColor = MaterialTheme.colorScheme.surface,
                 containerColor = Color(0x00FFFFFF),
```
nikhen-s, line 81: nit: Make sure to change this to primaryContainer when the Color and Material PR has been merged :)

app/src/main/java/com/github/se/eventradar/ui/event/EventDetails.kt
```diff
@@ -30,210 +30,137 @@ import androidx.compose.ui.text.font.FontWeight
...
-                .wrapContentSize()
-                .align(Alignment.Start)
-                .testTag("backButton"),
-            colors =
+fun EventDetails(navigationActions: NavigationActions) {
+
+  val fieldTitleColor = MaterialTheme.colorScheme.onSurfaceVariant
+  val fieldContentColor = MaterialTheme.colorScheme.onSurface
+
+  // Temporary text field values to be replaced with some ViewModel.stateVar.collectAsState()
```
nikhen-s, line 52: Nice!

app/src/main/java/com/github/se/eventradar/ui/event/EventDetails.kt
```diff
@@ -30,210 +30,137 @@ import androidx.compose.ui.text.font.FontWeight
...
+        onTabSelected = { tab -> navigationActions.navigateTo(tab) },
+        tabList = TOP_LEVEL_DESTINATIONS,
+        selectedItem = TOP_LEVEL_DESTINATIONS[0])
+  }
 }
 
-
 @Preview(showBackground = true)
 @Composable
 fun EventDetailsPreview() {
```
nikhen-s, line 159: nit: don't forget to remove this once the feature is implemented, Previews are generally only used during development :)

nikhen-s (changes requested):
Overall LGTM. Just a few nitpicks :)

Nath-Assa (commented):
LGTM!
Note: The EventDetails file currently contains hard-coded values. For maintainability and scalability, these should be replaced with dynamic data fetched from the Event list. This update is scheduled for Sprint 4.

Nath-Assa (approved):
LGTM!
Note: The EventDetails file currently contains hard-coded values. For maintainability and scalability, these should be replaced with dynamic data fetched from the Event list. This update is scheduled for Sprint 4.

app/src/main/java/com/github/se/eventradar/ui/event/EventDetails.kt
jgeaso1266: I personally don't see a problem with keeping the preview. They can also be useful for future developers who touch this code.

nikhen-s: Thanks for the comment, this makes sense and I agree too, I thought that Previews could slow down the app, but since our app doesn't consume too much resources right now I think it's ok + useful for future developers.

app/src/main/java/com/github/se/eventradar/ui/event/EventDetails.kt
```diff
@@ -0,0 +1,150 @@
...
+            Text(text = headerTime, color = fieldTitleColor, fontSize = contentTextSize)
+            Text(text = contentTime, color = fieldContentColor, fontSize = contentTextSize)
+          }
+        }
+
+    // register button
+    FloatingActionButton(
+        onClick = { /*TODO*/},
+        modifier = Modifier.padding(bottom = 16.dp, end = 16.dp).align(Alignment.End),
+        containerColor = MaterialTheme.colorScheme.surfaceContainer,
```
nikhen-s, line 135: nit: I think the containerColor could be primaryContainer (as it is a key UI element) for consistency :)

app/src/main/java/com/github/se/eventradar/ui/event/EventDetails.kt
```diff
@@ -0,0 +1,150 @@
...
+    ) {
+      Icon(
+          painter = painterResource(id = R.drawable.ticket),
+          contentDescription = "register to event button",
+          modifier = Modifier.size(32.dp),
+          tint = MaterialTheme.colorScheme.primaryContainer,
```
nikhen-s, line 141: Important: I think the tint could be onPrimaryContainer. Upon pulling the code, I couldn't see the ticket Icon because the color was really similar to surfaceContainer. (For elements on top of a container, use onContainerName)

nikhen-s (changes requested):
Just 2 small changes in the FloatingActionButton so that the colour of the ticket Icon can be seen, other than that, LGTM!

Viccha18:
Writing the tests for this code is the task I'm currently working on, I will make another PR for it as soon as possible

nikhen-s (approved):
LGTM.

PR #67, opened 2024-04-11
Added color and material from figma mockup

Viccha18:
# Objective

Add the figma material design to Colors.kt and Theme.kt, according to this : https://www.figma.com/file/yO8KgaCa6F39yUdyPSfkpn/Material-3-Design-Kit-(Community)?type=design&node-id=49823-12142&mode=design&t=QQREwP4DR1RFVCmz-0

### DIFF
--- a/app/src/main/java/com/github/se/eventradar/ui/theme/Color.kt
+++ b/app/src/main/java/com/github/se/eventradar/ui/theme/Color.kt
@@ -2,10 +2,62 @@ package com.github.se.eventradar.ui.theme
 
 import androidx.compose.ui.graphics.Color
 
-val Purple80 = Color(0xFFD0BCFF)
-val PurpleGrey80 = Color(0xFFCCC2DC)
-val Pink80 = Color(0xFFEFB8C8)
+// Primary colors
+val P_10 = Color(0xFF21005D)
+val P_20 = Color(0xFF381E72)
+val P_30 = Color(0xFF4F378B)
+val P_40 = Color(0xFF6750A4)
+val P_80 = Color(0xFFD0BCFF)
+val P_90 = Color(0xFFEADDFF)
+val P_100 = Color(0xFFFFFFFF)
 
-val Purple40 = Color(0xFF6650a4)
-val PurpleGrey40 = Color(0xFF625b71)
-val Pink40 = Color(0xFF7D5260)
+// Secondary colors
+val S_10 = Color(0xFF1D192B)
+val S_20 = Color(0xFF332D41)
+val S_30 = Color(0xFF4F378B)
+val S_40 = Color(0xFF625B71)
+val S_80 = Color(0xFFCCC2DC)
+val S_90 = Color(0xFFE8DEF8)
+val S_100 = Color(0xFFFFFFFF)
+
+// Tertiary colors
+val T_10 = Color(0xFF31111D)
+val T_20 = Color(0xFF492532)
+val T_30 = Color(0xFF633B48)
+val T_40 = Color(0xFF7D5260)
+val T_80 = Color(0xFFEFB8C8)
+val T_90 = Color(0xFFFFD8E4)
+val T_100 = Color(0xFFFFFFFF)
+
+// Error colors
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/se/eventradar/ui/theme/Color.kt
```diff
@@ -2,10 +2,57 @@ package com.github.se.eventradar.ui.theme
 
 import androidx.compose.ui.graphics.Color
 
-val Purple80 = Color(0xFFD0BCFF)
-val PurpleGrey80 = Color(0xFFCCC2DC)
-val Pink80 = Color(0xFFEFB8C8)
+val P_10 = Color(0xFF21005D)
```
christinehahaa, line 5: Code documentation/style wise, might be better to name these what the colour is, rather than just P_ e.g. Purple_10

app/src/main/java/com/github/se/eventradar/ui/theme/Theme.kt
christinehahaa, line 1: Looks good!

christinehahaa (changes requested):
Just a quick change to make documentation clearer - otherwise looks good!

app/src/main/java/com/github/se/eventradar/ui/theme/Color.kt
Viccha18: I agree, except that P do not stand for "purple" but for "primary". In the reference it is named like this and I did the same because looking up for each individual color name would be too time consuming. And it was also easier to then assign the color to the material

app/src/main/java/com/github/se/eventradar/ui/theme/Theme.kt
```diff
@@ -16,22 +16,76 @@ import androidx.compose.ui.platform.LocalView
 import androidx.core.view.WindowCompat
```
nikhen-s, line 17: Important: I think some comments would be helpful to describe the main important colours, what they mean/what they can be used for. This way, it is easier/faster to know how to use these colours & can save time on figuring it out.
E.g. 
// Primary colors
primary = P_80,                 // Primary color, often used for main UI elements.
onPrimary = P_20,               // Text color for content on top of primary background.
(these were just generated with ChatGPT btw)
Other than that, it looks good to me! Now we can use colours consistently, thank you!

nikhen-s (changes requested):
Overall, Looks Good To Me!
I think the main thing is just some comments in the Theme.kt, because that's what we will use for the colours.

Viccha18:
Thanks, I've added a bit of documentation. I hope this is clearer now !

app/src/main/java/com/github/se/eventradar/ui/theme/Theme.kt
```diff
@@ -15,23 +15,112 @@ import androidx.compose.ui.platform.LocalContext
...
+ *  Tertiary :        color used for additional accents and UI elements
+ *  Surface :         are the background colors for the whole screen
+ *  [...]Containers : are color for the containers containing primary, secondary or tertiary elements
+ *  on[...] :         are used for components situated on a component of the same color name without the "on"
+ *                    Eg : For text on a "primary color background" you give it the "onPrimary" color
+ *
+ *  If you have any doubts, you can find examples of UI and look at their material here :
+ *       https://www.figma.com/file/yO8KgaCa6F39yUdyPSfkpn/Material-3-Design-Kit-(Community)?type=design&node-id=53923-27456&mode=design&t=ir0UAE4XpwgRJMYn-0
+ * */
+
```
nikhen-s, line 30: Comments look great, well done!

nikhen-s (approved):
LGTM, nice comments :)

app/src/main/java/com/github/se/eventradar/ui/theme/Color.kt
christinehahaa: Makes sense, thanks for the explanation :)

christinehahaa (approved):
LGTM

PR #75, opened 2024-04-12
Tests for login

christinehahaa:
3 tests for login
- 1 x ui display
- 1 x successful login
- 1 x unsuccessful login (in progress)


app/src/androidTest/java/com/github/se/eventradar/login/LoginTest.kt
Nath-Assa, line 1: Great start on the LoginTest class. After completing the commented-out test for failed login scenarios, we can expand tests to verify navigation and asynchronous operations post-login to provide a more robust validation of the login functionality.

Nath-Assa (approved):
LGTM so far! During Sprint 4, consider completing the commented-out test and adding a few additional tests to verify navigation and asynchronous operations post-login. 
Also, this PR is still a draft. Needs to be changed to an Issue.

app/src/androidTest/java/com/github/se/eventradar/login/LoginTest.kt
```diff
@@ -0,0 +1,99 @@
...
+        performClick()
+      }
+
+      // assert that an Intent resolving to Google Mobile Services has been sent (for sign-in)
+      intended(toPackage("com.google.android.gms"))
+    }
+    Intents.release()
+  }
+
+  // TODO: Add test for failed log-in
```
jgeaso1266, line 51: We will have to discuss with the coaches if such a test is feasible.

nikhen-s: Yes, I think we should discuss it with the coaches, maybe the login could fail when there is no internet connection or when Firebase is down? 

The times when login failed for me are: when the firebase SHA1/google-services.json config file/or when the web-client id in the app was different.

jgeaso1266:
Sufficient tests were added in #171

PR #80, opened 2024-04-12
Initial Implementation of Home Screen UI with Mock Data and Assets

Nath-Assa:
In this pull request, I've laid the foundational UI elements for the Event Radar application's Home Screen. 

**### Assets Added:**

- Added the Event Radar logo and a placeholder image for events to the drawable directory to be used within the UI.

**Data Class:**

- Created an Event data class that will serve as the model for event information displayed in the UI.

**Home Screen Implementation (Home.kt):**

**Implemented basic features of the Home Screen which include:**

- A TabRow for "Browse" and "Upcoming" events. The "Browse" tab is functional, showing a list of events, while the "Upcoming" tab is currently a placeholder.

- An EventCard list that displays mock events. This serves as a prototype to visualize how the events will be presented to the users.

**Future Enhancements:**

- The bottom navigation bar needs to be added along with proper routing for the app.

- An EventViewModel should be created for handling the business logic, including querying the Events database.

- A search bar and filter button will be incorporated for better user experience in finding specific events.

- The map view button will be added to offer a geographic perspective of events.


app/src/main/res/drawable/event_logo.png
achillet3, line 1: The race is on between our PRs to see who manages to merge the new logo and place holder first hahah

app/src/main/java/com/github/se/eventradar/model/event/Event.kt
```diff
@@ -0,0 +1,23 @@
...
+
+data class Event(
+    val id: String,
+    val hostUserId: String,
+    val name: String,
+    val description: String,
+    val date: LocalDate,
+    val time: LocalTime,
+    val category: EventCategory
+    // TODO: Add 'location' attribute once the Map feature is implemented
```
achillet3, line 22: I would still add the location attribute even without the Map. Worst-case it's a string of the postal address that will be changed later.

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
```diff
@@ -0,0 +1,190 @@
...
+                style =
+                    TextStyle(
+                        fontSize = 19.sp,
+                        lineHeight = 17.sp,
+                        fontFamily = FontFamily(Font(R.font.roboto)),
+                        fontWeight = FontWeight(500),
+                        color = MaterialTheme.colorScheme.onPrimaryContainer,
+                        textAlign = TextAlign.Center,
+                        letterSpacing = 0.25.sp,
+                    ))
```
achillet3, line 105: To give the "Browse" text a little bit more breathing room I would add some padding here.
```suggestion
                    ))
                Spacer(modifier = Modifier.height(8.dp))
```
Note that this suggestion is based on the current implementation but in the future it should be done with a constraint layout.

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
```diff
@@ -0,0 +1,190 @@
...
```
achillet3, line 119: Same should be done here

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
```diff
@@ -0,0 +1,190 @@
...
+
+@Composable
+fun EventList(events: List<Event>) {
+  LazyColumn { items(events) { event -> EventCard(event) } }
+}
+
+@Composable
+fun EventCard(event: Event) {
+  Card(
+      modifier = Modifier.padding(horizontal = 16.dp, vertical = 8.dp).height(IntrinsicSize.Min),
```
achillet3, line 149: Here I would use a constraint layout to fix the left and right of the cards to the left and right, respectively, of the `TabRow`. 
```suggestion
      modifier = Modifier.padding(horizontal = 32.dp, vertical = 8.dp).height(IntrinsicSize.Min),
```
A temporary quick fix could be simply having the same horizontal padding as the `TabRow`

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
achillet3, line 1: Note for further improvements in functionality: Limit the maximum size of the cards (meaning limit the characters displayed from the description) so that there is still an acceptable number of events on the screen.

achillet3 (commented):
In general, this homepage looks very nice. Good use of the color Scheme! As currently developed, it is all ready to be merged into the master branch and used as our Homepage. However, there are a few recommendations in terms of design and functionality that I would recommend.

First, use a Constraint Layout to have a more flexible design especially as the number of components will multiply in this screen in the future. There is an example below of a good implementation that leverages constraints to have a better design.

Then, maybe for the next sprint, but I think it will be important to fix the maximum size of the cards to limit the amount of description displayed in this browser. This could either be through a maximum height of each card, leading to a limit of characters of the description being displayed, or we could add another attribute being a short description with a limit of characters and the full description will only be available in the event's details

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
wisdomseeker2001: Hi, so upon reviewing, if you read the blueprint, it is not the description of the event that should come up, this only once you click on the event. Please refer to 2.1 or 2.11. (only Pic, name, date and location should be there) - this is also based on many over event apps we have compared to.

achillet3: Good to know, will read the blueprint once more. If it is in the blueprint then please change the wireframes and mockups in Figma. They remain the point of reference when developing. Also looking at the actual implementation from @Nath-Assa the description is displayed for the moment so I believe the simplest way to achieve a functional implementation for tonight is to build upon what is here. Note: this improvement is rather a feature to be added in the future sprints as we are short in time...

achillet3 (commented):
Official review of this Pull Request, my previous review and comments still stand they were simply done on a draft Pull Request

achillet3 (approved):
All my aesthetic comments have been addressed. The location will be addressed in a future sprint. So all good for me. Well done!

Nath-Assa:
Thank you for the insightful comments! Will implement the changes in Sprint 4.

PR #112, opened 2024-04-15
Data model for Location and Event

wisdomseeker2001:
Hi, in this Pull Request:

-I finalised the Event.Kt data structure
-I also created a ticket data structure 
-I create a Location.kt file with the Location Data Structure
- I imported the FireBase Storage to the gradle 
- I changed home.kt such that it is compatible with the new Event structure


app/src/main/java/com/github/se/eventradar/model/Location.kt
```diff
@@ -0,0 +1,11 @@
...
+
+data class Location(
+    var latitude: Double,
+    var longitude: Double,
+    var address: String,
+)
+
+fun getLocation(name: String, lat: Double, lng: Double): Location {
+  return Location(lat, lng, name)
+}
```
Viccha18, line 11: nit: function name could be clearer, at first glance I didn't understand it was to create a new Location instance from values. I think "LocationFrom()" or "CreateLocation()" or any idea you have.

app/build.gradle.kts
```diff
@@ -140,6 +141,9 @@ dependencies {
     implementation(platform("com.squareup.okhttp3:okhttp-bom:4.12.0"))
     implementation("com.squareup.okhttp3:okhttp")
     implementation("com.squareup.okhttp3:logging-interceptor")
+
+    //QR CODE
+    implementation ("com.google.zxing:core:3.4.1")
 }
```
Viccha18, line 147: this shouldn't be in this PR (only in the one you made for the QR code generation)

app/src/main/java/com/github/se/eventradar/model/event/Event.kt
```diff
@@ -11,13 +11,19 @@ enum class EventCategory {
...
+    val location: Location,
     val description: String,
-    val date: LocalDate,
-    val time: LocalTime,
-    val category: EventCategory
-    // TODO: Add 'location' attribute once the Map feature is implemented
+    val ticket: Ticket,
+    val contact: String,
+    val organiserList: Set<String>,
+    val attendeeList: Set<String>,
```
Viccha18, line 26: I read in the blueprint that all coorganizer can edit the events they coorganize, can they as well delete it ?
if yes we can keep it like that, else we may require a "creator" value.

the type is a Set of string. Does it contains the users uid ? Could you specify it with a comment ?

app/src/main/java/com/github/se/eventradar/model/event/Event.kt
```diff
@@ -11,13 +11,19 @@ enum class EventCategory {
...
 }
 
+data class Ticket(val name: String, val price: Double, val quantity: Int)
+// new event struct final
 data class Event(
-    val id: String,
-    val hostUserId: String,
-    val name: String,
+    val eventName: String,
+    val eventPhoto: String,
```
Viccha18, line 18: nit: what does that String contains about the image ? a comment would be nice !

app/src/main/java/com/github/se/eventradar/model/event/Event.kt
```diff
@@ -11,13 +11,19 @@ enum class EventCategory {
...
-    val date: LocalDate,
-    val time: LocalTime,
-    val category: EventCategory
-    // TODO: Add 'location' attribute once the Map feature is implemented
+    val ticket: Ticket,
+    val contact: String,
+    val organiserList: Set<String>,
+    val attendeeList: Set<String>,
+    val category: EventCategory,
+    val fireBaseID: String
```
Viccha18, line 28: nit: I'd go for `val uid: String`, because it is shorter and I don't think there is the need to specify its for firebase.

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
```diff
@@ -207,7 +225,7 @@ fun EventCard(event: Event) {
                       .weight(1f),
               verticalArrangement = Arrangement.Center) {
                 Text(
-                    text = event.name,
+                    text = event.eventName,
```
Viccha18, line 228: nit : have "name" instead of "eventName" for the attribute to avoid the redondancy
(same goes for eventPhoto)

Viccha18 (changes requested):
This is great thank you ! I only have small remarks.

app/src/main/java/com/github/se/eventradar/model/event/Event.kt
```diff
@@ -11,13 +11,19 @@ enum class EventCategory {
   COMMUNITY
 }
 
+data class Ticket(val name: String, val price: Double, val quantity: Int)
```
jgeaso1266, line 14: I see that the `quantity` parameter is in the blueprint, but I remember us not coming to a consensus on this because 1) it adds complexity in the QR code scanning process and 2) it was an extra feature. Did we decide as a group to keep this? The blueprint does not mention how the quantity is used.

app/src/main/java/com/github/se/eventradar/model/event/Event.kt
jgeaso1266: I agree that this is an important distinction to figure out now. I also think it would make more sense to save the reference to the host's user ID in the User Database than just to have their contact email.

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
```diff
@@ -44,41 +45,58 @@ import androidx.compose.ui.unit.dp
...
 import com.github.se.eventradar.model.event.Event
 import com.github.se.eventradar.model.event.EventCategory
+import com.github.se.eventradar.model.event.Ticket
 import com.github.se.eventradar.ui.BottomNavigationMenu
 import com.github.se.eventradar.ui.navigation.TOP_LEVEL_DESTINATIONS
-import java.time.LocalDate
-import java.time.LocalTime
+import java.time.LocalDateTime
 
+@SuppressLint("SuspiciousIndentation")
```
jgeaso1266, line 56: This will fail the CI test. You will have to format your code to be in line with KTFMT to merge this into main.

jgeaso1266 (changes requested):
There are a few (small) things that are in the blueprint that I think we should be reviewed after looking at this implementation.

app/src/main/java/com/github/se/eventradar/model/event/Event.kt
Viccha18: I think this ticket quantity represent the max number of people that can attend the event and not the number of ticket one person can buy. So it should not complexify the scanning or I am forgetting something ?

app/src/main/java/com/github/se/eventradar/model/Location.kt
wisdomseeker2001: okay done

app/build.gradle.kts
wisdomseeker2001: sorry, accidental, I will swap

app/src/main/java/com/github/se/eventradar/model/event/Event.kt
wisdomseeker2001: For now, let us keep it this way, if we wanted to make the difference then we should implement the employee list and organiser list but decided not to and to keep it as an extra.
String in the list will be the uids of the user in database only :)

app/src/main/java/com/github/se/eventradar/model/event/Event.kt
wisdomseeker2001: string is reference to image path

app/src/main/java/com/github/se/eventradar/model/event/Event.kt
wisdomseeker2001: do not understand

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
wisdomseeker2001: I think eventName is better personally just to not get confused when used somewhere else

app/src/main/java/com/github/se/eventradar/model/event/Event.kt
wisdomseeker2001: Quantity will be how many tickets are available, 
every time one is purchased it will decrease by 1, at 0 => sold out

we are still keeping: 1 event purchase per user :)

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
wisdomseeker2001: done

app/src/main/java/com/github/se/eventradar/model/event/Event.kt
Viccha18: ok, doesn't matter

app/src/main/java/com/github/se/eventradar/model/event/Event.kt
Viccha18: Which means all users in the organisers list got the same right to edit and delete the event ?

Nath-Assa (approved):
LGTM! Make sure to resolve conflicts

Viccha18 (approved):
LGTM

Nath-Assa:
After reviewing the changes and the current state of the project, it appears that this pull request does not contribute any necessary modifications or enhancements. All intended updates are already in main.

PR #120, opened 2024-04-15
qrcode generator

wisdomseeker2001:
Qr code generation upon signup + Adding to FireStore + Adding Field Qr Code to the user in Firebase

This branch branches off main, has no dependencies on my previous PRs.

the function qrCodeProcess() needs to be called upon signup and everything is done automatically.
I created a Folder in the storage named QrCode where we will store all Qr Codes

@achillet3 - when testing with sample Users to see if (QR Generation successful + Upload to Firestore Successful + Added to User Qr Code Field Successful) - please add the appropriate field named "QrCode" to the sample Users the FireBase otherwise that part of code (line 66) does not make sense

The code is commented for guidance

### DIFF
--- a/app/src/main/java/com/github/se/eventradar/qrCode/QrCodeGenerator.kt
+++ b/app/src/main/java/com/github/se/eventradar/qrCode/QrCodeGenerator.kt
@@ -1,0 +1,82 @@
+package com.github.se.eventradar.qrCode
+
+import android.graphics.Bitmap
+import android.graphics.Color
+import com.google.firebase.firestore.FirebaseFirestore
+import com.google.firebase.storage.FirebaseStorage
+import com.google.zxing.BarcodeFormat
+import com.google.zxing.EncodeHintType
+import com.google.zxing.qrcode.QRCodeWriter
+import java.io.ByteArrayOutputStream
+import java.util.*
+
+class QrCodeGenerator {
+
+  fun qrCodeProcess(db: FirebaseFirestore, userId: String) {
+    uploadBitmapToFirestore(generateQRCode(userId), userId, db)
+  }
+
+  private fun generateQRCode(userId: String, width: Int = 512, height: Int = 512): Bitmap {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/se/eventradar/qrCode/QrCodeGenerator.kt
```diff
@@ -0,0 +1,82 @@
...
+import com.google.firebase.storage.FirebaseStorage
+import com.google.zxing.BarcodeFormat
+import com.google.zxing.EncodeHintType
+import com.google.zxing.qrcode.QRCodeWriter
+import java.io.ByteArrayOutputStream
+import java.util.*
+
+class QrCodeGenerator {
+
+  fun qrCodeProcess(db: FirebaseFirestore, userId: String) {
```
jgeaso1266, line 15: A better workflow would be:

1. call `qrCodeProcess`
2. `qrCodeProcess` generates the bitmap and uploads to cloud storage
3. `qrCodeProcess` _returns_ the download link
4. This download link is then used in the user object to upload the user to firebase

This way you don't need to worry about the userId. Now this is truly a utility function and doesn't depend on the user.

app/src/main/java/com/github/se/eventradar/qrCode/QrCodeGenerator.kt
```diff
@@ -0,0 +1,82 @@
...
+    uploadBitmapToFirestore(generateQRCode(userId), userId, db)
+  }
+
+  private fun generateQRCode(userId: String, width: Int = 512, height: Int = 512): Bitmap {
```
jgeaso1266, line 19: The issue you linked to this also included tests. You should add a test for these functions.

jgeaso1266 (changes requested):

jgeaso1266:
Is this PR still necessary in light of #236?

PR #123, opened 2024-04-16
Upcoming view model

wisdomseeker2001:
exact duplicate of code for allEvents View Model I just did just add line 19 I have

.whereArrayContains("AttendeeList", uid)

where uid will be string of uid of user in question that will be passed as argument, 
so the AttendeeList Field in Event on Firebased must contain that uid to be added to the list

@Nath-Assa when designing the UI, you must have 2 screens:
1) If List Empty => "You have No Upcoming Events"
2) Else display event


Nath-Assa:
After consideration, we've decided not to proceed with creating a new ViewModel specifically for managing upcoming events. Instead, we will enhance the existing EventsOverviewViewModel by adding a function to handle upcoming events.

PR #125, opened 2024-04-18
VM for AllEventsOverview with getEvents function and associated data classes

christinehahaa:
Key changes:
- Modified the allEventsOverview ViewModel from Joyti's PR: https://github.com/swent-sp-2024-event-radar/event-radar-app/pull/122
- Cleaned up the getEvents function from Joyti's PR
- Created EventList data class, modifying the structure of the getEvents function for usage for filtering events and getting a singular event
- Moved EventCategory and EventTicket data classes from Event dataclass to new file for clarity

PR #126, opened 2024-04-18
Dynamic Event Details screen

Viccha18:
# Objective

Link the Joyti's viewModel for events details with the Event Details screen made in sprint 3 

# Detailed changes
**Event Details Screen UI**
- Made a separate file for the event UI component (will add in the futur the "editable" components for Create Event screen)
- Added Constraint Layout
- Removed static string in Text(), and replaced them with ViewModel EventUiState
- Added many string ressource used in the display of an event (Title words such as "description" "date"... & conversion for the EventCategory enum)

**Navigation**
- Added composable route to navGraph and take a "eventId" argument
- call navigate with Route.EVENT_DETAILS & eventId from Home.kt 

**EventCategory enum**
- Added a `toString(context: Context)` fonction to Event Category enum to get a res defined string to display
- Added "PARTY" ~and "UNDEFINED"~ in EventCategory enum

**EventDetailsViewModel**
- Removed initialzation function of the EventUiState dataclass which was not necessary
- Changed var to val in the EventUiState dataclass 
- ~Create instance of Event from fetched data. Was: (data from firebase --> EventUiState) Now: (data from firebase --> new Event --> EventUiState)~

**Tests**
- Created a mockViewModel in for the existing display test, and a sample EventUiState
- one test for EventDetailsViewModel

# Note 
~- TODO need to mock the firestore to test the viewModel in depth. 
Christine and I are thinking about making a "Database" class (like proposed in the bootcamp) that would contain the function that interact with the database and that takes care of converting the data from the firebase format to our own data model (Event, User, ....)~


app/src/main/java/com/github/se/eventradar/model/event/Event.kt
```diff
@@ -8,7 +10,21 @@ enum class EventCategory {
   SPORTS,
   CONFERENCE,
   EXHIBITION,
-  COMMUNITY
+  COMMUNITY,
+  PARTY,
+  UNDEFINED;
```
jgeaso1266, line 15: These my clash with #125

jgeaso1266 (changes requested):
Wow, big PR. A few important things that should be addressed, but mostly comments and questions.

app/src/main/java/com/github/se/eventradar/ui/event/EventDetails.kt
```diff
@@ -18,159 +16,43 @@ import androidx.compose.material3.MaterialTheme
...
-  }
-}
-
-@Composable
-fun EventDetails(navigationActions: NavigationActions) {
-
-  val fieldTitleColor = MaterialTheme.colorScheme.onSurfaceVariant
-  val fieldContentColor = MaterialTheme.colorScheme.onSurface
+  val componentStyle =
+      EventComponentsStyle(
```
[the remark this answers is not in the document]
Viccha18: Since I separted the "description "distance" "date & time" etc.. in different composable function I wanted to avoid
- having a tons of function parameters
- to hard code the color scheme in each component when they need to be harmonized

app/src/main/java/com/github/se/eventradar/ui/event/EventUiComponents.kt
```diff
@@ -0,0 +1,123 @@
...
+fun EventDistance(modifier: Modifier, eventUiState: EventUiState, style: EventComponentsStyle) {
+  Column(modifier = modifier) {
+    Text(
+        text = stringResource(id = R.string.event_distance),
+        style = style.subTitleStyle,
+        color = style.subTitleColor,
+        modifier = Modifier.testTag("distanceTitle"))
+    Text(
+        // TODO implement a function that convert event location to distance from user
+        // TODO require user data class
```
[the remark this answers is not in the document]
Viccha18: As we discussed, when a user browse events and clicks on the details, he don't see the location but at what distance the event is from him. Currently, the Location object contains geoPoint and address, so we still to have a function that takes those geo points and the position of the user to compute the distance between them

app/src/main/java/com/github/se/eventradar/model/event/Event.kt
```diff
@@ -8,7 +10,21 @@ enum class EventCategory {
...
+
+  fun toString(context: Context): String {
```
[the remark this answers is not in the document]
Viccha18: ok, I'll make some more research, or do you know how I can get it ? I tried stringResource() but it needed a composable function.

app/src/main/java/com/github/se/eventradar/model/event/EventDetailsViewModel.kt
```diff
@@ -13,76 +15,98 @@ import kotlinx.coroutines.flow.StateFlow
...
-    var location: Location,
-    var description: String,
-    var ticket: Ticket,
-    var contact: String,
-    var category: EventCategory,
+    val eventName: String = "",
+    val eventPhoto: String = "",
+    val start: LocalDateTime = LocalDateTime.MIN,
+    val end: LocalDateTime = LocalDateTime.MAX,
+    val location: Location = Location(0.0, 0.0, ""), // TODO add standard val to location class ?
```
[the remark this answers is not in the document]
Viccha18: Sorry, I should have read my comments before pushing. I meant a default value. One that we can use by default if there's problem with the database data (similar to the UNDEFINED EventCategory that I added in case the "category field" doesn't contain a valide category)

app/src/main/java/com/github/se/eventradar/model/event/EventDetailsViewModel.kt
```diff
@@ -13,76 +15,98 @@ import kotlinx.coroutines.flow.StateFlow
...
-            // Document does not exist
-            println("Document does not exist")
-          }
-        }
-        .addOnFailureListener { exception ->
-          // Handle any errors that occur during fetching
-          // You may want to log the error or show a message to the user
-          println("Error fetching event: $exception")
-        }
+  private fun createEventFromDocument(document: DocumentSnapshot, eventId: String): Event {
```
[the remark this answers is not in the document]
Viccha18: That's a good point. I like to separate things (maybe to much). I am wondering what the Event model will be used for. Because one of my next task is to create the Create event screen and add functions to the viewModel, and naturally later, there will be the Edit event Screen as well...

You're right that there is no need now to have them separate, so I can remove that middle man.

app/src/main/java/com/github/se/eventradar/model/event/Event.kt
Viccha18: Right ty, I will discuss it with Christine

RaulinN (commented):
Couldn't review the 100% thoroughly because I couldn't find a way to run the code on the emulator (didn't have any eventIds and clicking on an event from the homepage did not seem to work. Overall, from what I could see, that's some solid work well done!

wisdomseeker2001:
Hello,

I took over Vicky's Branch (https://github.com/swent-sp-2024-event-radar/event-radar-app/pull/126) and made a duplicate. It needed to be adapted to the new structure of the project, that being mainly a new class for all the firebase communication (FirebaseEventRepository.kt) and also the wonderful modifications that @jgeaso1266 made the Event.kt, making the code a lot more clean.

I only changed/ added two classes:

Modified EventDetailsViewModel.kt:
-Adapted it such that it calls the getEvent() function in FirebaseEventRepository.kt and then updated the Ui State.

Created EventOverviewViewmodelUnitTest.kt:
This is a really basic test just to see if the whole object was initialised correctly,
If it fails, I will then break down to assert each of the objects parameter's to see which one is not being initialised properly

I also added the sample event to the Firebase :)

all the other changes you see are not mine, I couldn't get them to go away... they have all even commented on in https://github.com/swent-sp-2024-event-radar/event-radar-app/pull/126

Please let me know if you see any errors

app/src/main/java/com/github/se/eventradar/model/event/EventCategory.kt
```diff
@@ -1,12 +1,28 @@
...
 enum class EventCategory(val displayName: String) {
   MUSIC("Music"),
   SPORTS("Sports"),
   CONFERENCE("Conference"),
   EXHIBITION("Exhibition"),
   COMMUNITY("Community"),
   SOCIAL("Social"),
+  PARTY("Party");
+
+  fun toString(context: Context): String {
```
jgeaso1266, line 15: this is already implemented by [Kotlin](https://stackoverflow.com/questions/28548015/how-do-i-create-an-enum-from-a-string-in-kotlin)

app/src/main/java/com/github/se/eventradar/ui/event/EventDetails.kt
```diff
@@ -18,159 +16,48 @@ import androidx.compose.material3.MaterialTheme
...
-        text = "start $contentTime",
-        style = standardTextStyle,
-        color = contentColor,
-        modifier = Modifier.testTag("timeStartContent"))
-    Text(
-        text = "end $contentTime",
-        style = standardTextStyle,
-        color = contentColor,
-        modifier = Modifier.testTag("timeEndContent"))
+  LaunchedEffect(Unit) { // Using `Unit` as a key to run only once
```
jgeaso1266, line 49: Consider putting this in the `init` block of the view model instead

app/src/main/java/com/github/se/eventradar/ui/event/EventDetails.kt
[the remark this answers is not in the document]
jgeaso1266: Ok, make sure this is in line with what the rest of the app looks like.

app/src/main/java/com/github/se/eventradar/viewmodel/EventDetailsViewModel.kt
```diff
@@ -1,91 +1,68 @@
...
-    private val db: FirebaseFirestore = Firebase.firestore,
-    eventId: String
+@HiltViewModel
+class EventDetailsViewModel
+@Inject
+constructor(
+    private val eventRepository: IEventRepository,
 ) : ViewModel() {
 
-  private val _uiState = MutableStateFlow(initialEventState())
```
jgeaso1266, line 21: These shouldn't be deleted...

jgeaso1266 (commented):
Good overall but a few questions and suggestions.

app/src/main/java/com/github/se/eventradar/model/event/EventCategory.kt
Viccha18: I know there already exist a function from kotlin that will convert the enum value to string, but the idea (maybe I think touo far) is to map each type of the enum to a Ressource String, which as I understood is of use to easily translate the app and its content. Since I used string ressource for subtitle such as "description", "location" I wanted to stay consistent and have string ressource to display the event Category as well

app/src/main/java/com/github/se/eventradar/viewmodel/EventDetailsViewModel.kt
```diff
@@ -1,91 +1,68 @@
...
+                category = response.data.category,
+            )
           }
         }
-        .addOnFailureListener { exception ->
-          // Handle any errors that occur during fetching
-          // You may want to log the error or show a message to the user
-          println("Error fetching event: $exception")
-        }
+        is Resource.Failure -> Log.d("EventDetailsViewModel", "Error getting event: ")
```
Nath-Assa, line 52:
```suggestion
       is Resource.Failure -> Log.e("EventDetailsViewModel", "Error getting event: ${response.throwable.message}")
```

app/src/test/java/com/github/se/eventradar/EventDetailsViewmodelUnitTest.kt
Nath-Assa, line 1: Consider expanding the test coverage for EventDetailsViewModel to account for a variety of scenarios. To begin, you can verify that all properties of EventUiState are correctly updated after data retrieval. Additionally, it would be beneficial to implement tests that examine how the ViewModel handles scenarios where the event data is modified or removed from the database, ensuring the model reacts appropriately to data changes or unavailability.

Nath-Assa (commented):
Nice work! I made some suggestions. I would also like you to add a few more tests to the ViewModel before I approve.

app/src/main/java/com/github/se/eventradar/viewmodel/EventDetailsViewModel.kt
Nath-Assa: If the `toString(context: Context)` method in the EventCategory enum simplifies the usage of localized string resources, I recommend keeping it. However, to enhance readability and avoid confusion with the standard toString() method, consider renaming it to something more descriptive like `getLocalizedName(context: Context)`

app/src/main/java/com/github/se/eventradar/model/event/EventCategory.kt
```diff
@@ -1,18 +1,36 @@
...
 enum class EventCategory(val displayName: String) {
   MUSIC("Music"),
   SPORTS("Sports"),
   CONFERENCE("Conference"),
   EXHIBITION("Exhibition"),
   COMMUNITY("Community"),
   SOCIAL("Social"),
+  PARTY("Party");
+
+  fun convertToResString(context: Context): String {
```
jgeaso1266, line 16: Still don't think this is necessary. It adds more complexity to the implementation when one can just use the enum.

app/src/main/java/com/github/se/eventradar/ui/event/EventDetails.kt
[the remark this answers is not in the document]
jgeaso1266: Is the color the same as the rest of the app? I don't think we have implemented specific colors like you have (we are currently just using default colors). If that is the case, then maybe take this out of this PR for now so it can be uniform for M2.

app/src/main/java/com/github/se/eventradar/viewmodel/EventDetailsViewModel.kt
jgeaso1266: > `initialEventState()` which was used to initialized the UiState does not exist anymore (initialization is now done directly in the dataclass) maybe you're talking about
> 
> ```kotlin
>   private val _uiState = MutableStateFlow(EventUiState())
>   val uiState: StateFlow<EventUiState> = _uiState
> ```
> 
> at line 31&32

I didn't see this. For visibility and consistency with other view models, this should be the first two lines of the view model.

jgeaso1266 (commented):
A few questions. I am not sure if they require fixes so not going to "request changes" but will approve once you answer the questions.

app/src/main/java/com/github/se/eventradar/model/event/EventCategory.kt
Viccha18: I cannot reply under your other comments so I'll answer here

1. I can agree that it is of no use in V1 of our app. But as a french speaker, I would like to have the app in french lol, and we wouldn't put every translation in the enum. This is why I implemented this using the resource string. But if you're strongly against having that function in V1, I can remove it.

2. Colors are exactly the same as the one already used in the rest of the app. I use the default colorScheme with "onSurface" "onSurface Variant" etc... I just didn't hard coded the color scheme in my individual UI components. We can discuss the relevance of this. Because at that point in the app, colorScheme colors could be hard coded in the components since they always take the same primary material and we have currently no use to switch them to secondary color scheme.

jgeaso1266 (approved):
LGTM

PR #129, opened 2024-04-18
Replace mock events with events from Firebase

jgeaso1266:
## Objective
- Link the `EventsOverviewViewModel` to the Home screen
- query events from the database into browse events

## Key Changes
- [ ] remove `mockEvents`
- [ ] use the view model `getEvents()` function to get events from Firebase
- [ ] add a test


Viccha18 (approved):
LGTM

Issue #110, opened 2024-04-15
Create "My Hosting" page: UI/View + Tests


nikhen-s:
## Link To User Story and Figma
User Story #76, Figma Issue #98 

## Objective
- Implement the UI upon clicking My Hostings in Bottom Navigation Bar

## Acceptance Criteria
- [ ] Contains "+" Button to create new event.
- [ ] If user does not host an event, displays text "No hosted events"
- [ ] Displays events that the user hosts in cards containing the event details. Upon clicking, it leads to event overview.

christinehahaa:
Task was done in sprint 4 (XXX hrs @nikhen-s to update) and is to be finalized in sprint 5.

PR #127, opened 2024-04-18
Add Map View to Browse Events

jgeaso1266:
## Objective
- be able to view the approximate location of all events on a Map
- achieve feature parity with the list view

## Key Changes
- [ ] Created `Map.kt` that holds google maps composable
- [ ] link the `EventsOverviewViewModel()` to the Map UI
- [ ] Added floating action button to home screen
- [ ] Added logic to show the correct list view (map or list)
- [ ] Added UI test

closes #68 
closes #118

PR #130, opened 2024-04-18
My_Hosting UI + tests

nikhen-s:
Related to issue: https://github.com/swent-sp-2024-event-radar/event-radar-app/issues/110 ==> closes https://github.com/swent-sp-2024-event-radar/event-radar-app/issues/110

## Objective
- Implement the UI upon clicking My Hostings in Bottom Navigation Bar

## Acceptance Criteria
- [x] Contains "+" Button to create new event. 
- [x] Clicking party icon in Nav Bar leads to My_Hosting UI
- [x] Created tests that verified elements are displayed correctly
- [x] Implemented HostedEventsViewModel

## Changes
- Added Tests for ViewModel and UI for more coverage
- Completely Separated the State from UI, it is now all in ViewModel
- Created a `ReusableComponents` class for code-reusability
- Added test tags to searchBar to assertDisplay
- Added Functions in EventOverviewViewModel and HostedEventsViewModel to update viewList/Tab change status.


app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -0,0 +1,281 @@
...
+                .testTag("logo"),
+        verticalAlignment = Alignment.CenterVertically) {
+          Image(
+              painter = painterResource(id = R.drawable.event_logo),
+              contentDescription = "Event Radar Logo",
+              modifier = Modifier.size(width = 186.dp, height = 50.dp))
+        }
+
+    TabRow(
+        selectedTabIndex = selectedTabIndex,
```
RaulinN, line 126: We are using TabRows already in the Home screen. Maybe you could create a component out of these tabs in order not to repeat a large amount of code across the app?

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -0,0 +1,281 @@
...
+              "Join the Community",
+              EventTicket("regular", 0.0, 10000),
+              "valerian@joytigoel.com",
+              mutableSetOf("298jhk", "jwj8223"),
+              mutableSetOf("20982e2hk", "j1ou223e8223"),
+              EventCategory.COMMUNITY,
+              "89298"))
+
+  var selectedTabIndex by remember { mutableIntStateOf(0) }
+  val context = LocalContext.current
```
RaulinN, line 106:
```suggestion
```

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -0,0 +1,281 @@
...
+                  top.linkTo(parent.top, margin = 32.dp)
+                  start.linkTo(parent.start, margin = 16.dp)
+                }
+                .testTag("logo"),
+        verticalAlignment = Alignment.CenterVertically) {
+          Image(
+              painter = painterResource(id = R.drawable.event_logo),
+              contentDescription = "Event Radar Logo",
+              modifier = Modifier.size(width = 186.dp, height = 50.dp))
+        }
```
RaulinN, line 123: The logo should also be a component. Every time the logo is used on a view, there's code duplication

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
RaulinN: Additionally, if you don't plan on using a second tab (there's currently one), the `TabRow` + `Tab` should be deleted entirely

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -0,0 +1,281 @@
...
+
+    if (selectedTabIndex == 0) { // events I host vs events I staff?
+      EventList(
+          mockEvents,
+          Modifier.fillMaxWidth().constrainAs(eventList) {
+            top.linkTo(tabs.bottom, margin = 8.dp)
+            start.linkTo(parent.start)
+            end.linkTo(parent.end)
+          })
```
RaulinN, line 165: Reusing `EventList` is great! I would move the `EventList` composable from the `Home` file into a `component` folder. It is cleaner this way than importing stuff from Home.kt into Hosting.kt. This might also lead to cyclic dependencies which we want to avoid at all costs

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -0,0 +1,281 @@
...
+                }
+                .testTag("floatingActionButtons")
+                .fillMaxWidth()
+                .padding(16.dp),
+        horizontalArrangement = Arrangement.Absolute.Left,
+        verticalAlignment = Alignment.CenterVertically) {
+          ExtendedFloatingActionButton(
+              text = {
+                Text(
+                    text = "Create New Event",
```
RaulinN, line 183: Make sure you have your strings in strings.xml file

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -0,0 +1,281 @@
...
+        onTabSelected = { tab -> navigationActions.navigateTo(tab) },
+        tabList = TOP_LEVEL_DESTINATIONS,
+        selectedItem = TOP_LEVEL_DESTINATIONS[3],
+        modifier =
+            Modifier.testTag("bottomNavMenu").constrainAs(bottomNav) {
+              bottom.linkTo(parent.bottom)
+              start.linkTo(parent.start)
+              end.linkTo(parent.end)
+            })
+  }
```
RaulinN, line 225: You (as a group) are duplicating a lot of code! This snippet (modulo the selectedItem index) is already present 3 times in the code
- short fix (not recommended) : Create a component that represents this bottom navigation bar that you can use from all the top level navigation screens (e.g. Home, MyHosting, Profile)
- recommended fix : you should not have to hardcode the index of the `selectedItem`. Can you think of a way to remove this information?

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -0,0 +1,281 @@
...
+}
+
+@Composable
+fun EventList(events: List<Event>, modifier: Modifier = Modifier) {
+  LazyColumn(modifier = modifier) { items(events) { event -> EventCard(event) } }
+}
```
RaulinN, line 231: Ahhh... Nevermind then. So the `EventList` is imported from Home, but also redefined (same for EventCard). There's a """huge""" problem of code duplication in your project guys! Pay attention to what your teammates write, and use the features already implemented instead of copy-pasting the same code in a diffeerent file!

RaulinN:
Note, you can close the associated issue (#110 in this case) automatically when merging the PR. Make sure you use one of these special keywords in the PR description : https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

TLDR: "Related to issue: #110" ==> "closes #110" in your description

app/src/main/java/com/github/se/eventradar/ui/component/ReusableComponents.kt
```diff
@@ -0,0 +1,161 @@
...
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import com.github.se.eventradar.R
+import com.github.se.eventradar.model.Location
+import com.github.se.eventradar.model.event.Event
+import com.github.se.eventradar.model.event.EventCategory
+import com.github.se.eventradar.model.event.EventTicket
+import java.time.LocalDateTime
+
+fun getMockEvents(): List<Event> {
```
jgeaso1266, line 46: This is no longer necessary.

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -0,0 +1,155 @@
...
+import com.github.se.eventradar.R
+import com.github.se.eventradar.ui.BottomNavigationMenu
+import com.github.se.eventradar.ui.component.*
+import com.github.se.eventradar.ui.navigation.NavigationActions
+import com.github.se.eventradar.ui.navigation.TOP_LEVEL_DESTINATIONS
+import com.github.se.eventradar.ui.navigation.getTopLevelDestination
+
+@Composable
+fun HostingScreen(navigationActions: NavigationActions) {
+  val mockEvents = getMockEvents()
```
jgeaso1266, line 45: You should be able to directly use the event repository.

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -0,0 +1,155 @@
...
+    if (viewMapOrListIndex == 0) {
+      EventList(
+          mockEvents,
+          Modifier.fillMaxWidth().constrainAs(eventList) {
+            top.linkTo(divider.bottom, margin = 8.dp)
+            start.linkTo(parent.start)
+            end.linkTo(parent.end)
+          })
+    } else {
+      // To Do (Map View)
```
jgeaso1266, line 78: You should be able to implement this, Map is available as of #127.

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -0,0 +1,155 @@
...
+            Modifier.constrainAs(ref = buttons) {
+                  bottom.linkTo(bottomNav.top)
+                  centerHorizontallyTo(parent)
+                }
+                .testTag("floatingActionButtons")
+                .fillMaxWidth()
+                .padding(16.dp),
+        horizontalArrangement = Arrangement.Absolute.Left,
+        verticalAlignment = Alignment.CenterVertically) {
+          CreateEventFab(onClick = { /*TODO*/}, modifier = Modifier.testTag("createEventFab"))
```
jgeaso1266, line 92: If this is not working, you should put a Toast message to notify the user.

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -0,0 +1,155 @@
...
+              onClick = {
+                viewMapOrListIndex = if (viewMapOrListIndex == 0) 1 else 0
+                viewToggleIcon =
+                    if (viewMapOrListIndex == 0) Icons.Default.Place
+                    else Icons.AutoMirrored.Filled.List
+              },
+              iconVector = viewToggleIcon)
+        }
+    BottomNavigationMenu(
+        onTabSelected = { tab -> navigationActions.navigateTo(tab) },
```
jgeaso1266, line 107: nit: you can use `navigationActions::navigateTo` to avoid the lambda expression.

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -0,0 +1,155 @@
...
+              start.linkTo(parent.start)
+              end.linkTo(parent.end)
+            })
+  }
+}
+
+@Composable
+fun Title(modifier: Modifier) {
+  Text(
+      text = LocalContext.current.getString(R.string.hosted_events_title),
```
jgeaso1266, line 122: Not sure, but can you use `stringResource()` instead?

app/src/main/java/com/github/se/eventradar/ui/navigation/NavigationActions.kt
```diff
@@ -41,6 +42,16 @@ class NavigationActions(val navController: NavController) {
   }
 }
 
+fun getTopLevelDestination(targetTextId: Int): TopLevelDestination {
```
jgeaso1266, line 45: `Route` is a better/more reliable parameter than the text id, which can change at any time.

app/src/main/res/drawable/map.png
jgeaso1266, line 1: This icon should already be in the repo.

jgeaso1266 (changes requested):
In general, code is well written. I did not test on my device yet, so not sure how it looks. Some small notes.

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -0,0 +1,177 @@
...
+import com.github.se.eventradar.util.toast
+
+@Composable
+fun HostingScreen(
+    viewModel: EventsOverviewViewModel = viewModel(),
+    navigationActions: NavigationActions
+) {
+  val uiState by viewModel.uiState.collectAsState()
+  LaunchedEffect(key1 = uiState.eventList) { viewModel.getEvents() }
+  var viewMapOrListIndex by remember { mutableIntStateOf(0) }
```
jgeaso1266, line 58: nit: I remember @RaulinN saying something along the lines that these types of variables should go into the view model.

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -0,0 +1,177 @@
...
+import com.github.se.eventradar.ui.map.EventMap
+import com.github.se.eventradar.ui.navigation.NavigationActions
+import com.github.se.eventradar.ui.navigation.Route
+import com.github.se.eventradar.ui.navigation.TOP_LEVEL_DESTINATIONS
+import com.github.se.eventradar.ui.navigation.getTopLevelDestination
+import com.github.se.eventradar.util.toast
+
+@Composable
+fun HostingScreen(
+    viewModel: EventsOverviewViewModel = viewModel(),
```
jgeaso1266, line 53: You will want a separate view model for the Hosting screen. I have some comments later to show why.

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -0,0 +1,177 @@
...
+    navigationActions: NavigationActions
+) {
+  val uiState by viewModel.uiState.collectAsState()
+  LaunchedEffect(key1 = uiState.eventList) { viewModel.getEvents() }
```
jgeaso1266, line 57: This will return all events in the `events` database, you want it to only return the events hosted by the user.

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -0,0 +1,177 @@
...
+                    {
+                      top.linkTo(logo.bottom, margin = 16.dp)
+                      start.linkTo(parent.start)
+                      end.linkTo(parent.end)
+                    }))
+    HorizontalDivider(
+        modifier = Modifier.constrainAs(divider, { top.linkTo(title.bottom, margin = 10.dp) }))
+    if (viewMapOrListIndex == 0) {
+      EventList(
+          uiState.eventList
```
jgeaso1266, line 83: I see your comment, but `filteredEvent` is meant to filter the events on the homescreen, not the hosting screen.

jgeaso1266 (changes requested):
Your changes look mostly ok (just a couple of nits). But I see you are using the event overview view model when you should create a separate view model for the `Hosting` screen.

app/src/main/java/com/github/se/eventradar/viewmodel/HostedEventsViewModel.kt
Nath-Assa, line 1: For the HostedEventsUiState data class, I supposed the viewList field represents whether the user is in List View or Map view. However, we need to stay consistent. If we add this field in this UI state, we must also add it in EventsOverviewUiState. Additionally, would love to know your rationale behind storing this directly in the UiState rather than handling it Additionally, would love to know your rationale behind storing this directly in the UiState rather than handling it in Hosting.kt.

Nath-Assa, line 1: getHostedEvents() is very similar to getUpcomingEvents() which led to us exceeding the duplicate code threshold. Look into a way to modularize this, if possible.

Nath-Assa (changes requested):
Nice work on the code overall. However, this pull request seems quite extensive. I recommend dividing it into two separate pull requests for better manageability. Moreover, we should not proceed with merging this PR until the HostingEventsViewModel has been comprehensively tested.

app/src/androidTest/java/com/github/se/eventradar/hosting/HostingTest.kt
Nath-Assa, line 1: Ideally, we should have separate test files: one for UI interactions within the androidTest directory and another specifically for the HostedEventsViewModel logic in the test directory.

app/src/main/java/com/github/se/eventradar/model/EventsOverviewViewModel.kt
Nath-Assa, line 1: This file is not up to date with main. Please update from main:
Note: Since this file has been moved to the viewmodel directory, it should just be deleted.

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
Nath-Assa, line 1: You are directly accessing FirebaseAuth.getInstance().currentUser?.uid within the composable function. While this works, it's generally better to handle such data fetching or state management outside of the UI layer.

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -0,0 +1,179 @@
...
+import com.google.firebase.auth.FirebaseAuth
+
+@Composable
+fun HostingScreen(
+    viewModel: HostedEventsViewModel = hiltViewModel(),
+    navigationActions: NavigationActions
+) {
+  val uiState by viewModel.uiState.collectAsState()
+  val currentUser = FirebaseAuth.getInstance().currentUser?.uid
+  LaunchedEffect(key1 = uiState.eventList) { viewModel.getHostedEvents(currentUser.toString()) }
```
Nath-Assa, line 60: Please ensure that the case when `currentUser` is null (no user signed in) is safely handled, this will result in "null" being passed as a string, which might not be the intended behavior. Consider adding a null check or a default behavior when the user is not signed in.

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
```diff
@@ -126,7 +130,10 @@ fun HomeScreen(
         contentColor = MaterialTheme.colorScheme.primary) {
           Tab(
               selected = selectedTabIndex == 0,
-              onClick = { selectedTabIndex = 0 },
+              onClick = {
+                uiState.tab = Tab.BROWSE
```
jgeaso1266, line 134: You will not be able to update the uiState this way, you will have to have a helper function with access to `_uiState` to change it.

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
```diff
@@ -225,30 +236,25 @@ fun HomeScreen(
...
-    FloatingActionButton(
+    ViewToggleFab(
+        modifier =
+            Modifier.padding(16.dp).testTag("viewToggleFab").constrainAs(viewToggle) {
+              bottom.linkTo(bottomNav.top)
+              absoluteRight.linkTo(parent.absoluteRight)
+            },
         onClick = {
-          viewToggleBrowseIndex = if (viewToggleBrowseIndex == 0) 1 else 0
+          uiState.viewList = !uiState.viewList
```
jgeaso1266, line 253: Same here, to change the uiState, you will need access to `_uiState` in the viewModel. I see that there's no compilation error because you made the variable mutable (using `var`). It should not be possible to directly edit the UI state of the view model (it breaks the MVVM architecture). Instead, make everything a `val` and use helper functions to modify it. See examples in the bootcamp if needed. 

I see @christinehahaa also did this for her changes, so it would be good to sync up with her as well.

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -0,0 +1,179 @@
...
+import com.github.se.eventradar.ui.navigation.NavigationActions
+import com.github.se.eventradar.ui.navigation.Route
+import com.github.se.eventradar.ui.navigation.TOP_LEVEL_DESTINATIONS
+import com.github.se.eventradar.ui.navigation.getTopLevelDestination
+import com.github.se.eventradar.util.toast
+import com.github.se.eventradar.viewmodel.HostedEventsViewModel
+
+@Composable
+fun HostingScreen(
+    uid: String,
```
jgeaso1266, line 55: I might have been the one to recommend this, but after researching this myself, this won't work with the bottom navigation. #171 now allows for the correct user ID to be received using `FirebaseAuth.getInstance().currentUser!!.uid`.

app/src/main/java/com/github/se/eventradar/ui/navigation/NavGraph.kt
```diff
@@ -34,8 +36,14 @@ fun NavGraph(navController: NavHostController) {
...
     }
     composable(Route.MY_HOSTING) {
-      HomeScreen(navigationActions = navActions)
-      context.toast("My hosted events screen needs to be implemented")
+      val currentUser = FirebaseAuth.getInstance().currentUser
+      val uid =
+          if (currentUser != null) {
+            currentUser.uid
+          } else {
+            throw IllegalStateException("User is not authenticated")
```
jgeaso1266, line 44: Not a good idea to throw an uncaught exception in Android code. We can use null assertion instead. I also think it's better to have this call in the viewModel than in the navigation (but if you think this is better, I'd be curious to know you reasoning).

jgeaso1266 (changes requested):
A few important changes regarding the uiState, and a suggestion with the hosting screen. Also, you are at 69.2% coverage. There are only a few more lines that need to be covered looking at SonarCloud. Maybe invest a little time trying to get coverage to >=80%. Hopefully shouldn't take long.

app/src/main/java/com/github/se/eventradar/ui/navigation/NavGraph.kt
nikhen-s: I think that it's better for this call to be in viewModel too rather than NavGraph since only the eventCards rendered will depend on the userId, other parts of the Hosting UI itself should not depend on the userId, therefore it should be passed as a dependency only to the getHostedEvents function. The reason I passed it from navigation was to avoid calling user data from the HostingUI, but it didn't really solve the problem because I still called it in the NavGraph which is still part of UI layer. I think calling the user in the viewModel is a better way to do it such that it prepares all the data for the UI, instead of having the NavGraph do it. Thanks!

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
nikhen-s: I tested it with the bottom navigation and it worked without errors since the bottom navigation bar navigates to different screen based on the routes, and the initialization of uid is defined in the NavController to go to the HostingScreen composable. However, it makes sense to put the uid in ViewModel instead, so I will move it there (my reply to your comment below has my full reasoning to follow your suggestion).

app/src/main/java/com/github/se/eventradar/viewmodel/HostedEventsViewModel.kt
```diff
@@ -23,40 +25,51 @@ constructor(
...
-        is Resource.Failure -> {
-          Log.d("HostedEventsViewModel", "Error fetching user document")
-          _uiState.value =
-              _uiState.value.copy(eventList = EventList(emptyList(), emptyList(), null))
-        }
       }
     }
   }
+
+  fun onViewListStatusChanged(
```
jgeaso1266, line 67: nit: this doesn't need to take in the boolean. Whenever it's called, we know the new value will be the opposite of what it is now.

jgeaso1266 (commented):
One question and a few nits

app/src/main/java/com/github/se/eventradar/ui/component/ReusableComponents.kt
```diff
@@ -0,0 +1,303 @@
...
+import androidx.compose.ui.text.style.TextOverflow
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import com.github.se.eventradar.R
+import com.github.se.eventradar.model.event.Event
+import com.github.se.eventradar.model.event.EventCategory
+import com.github.se.eventradar.viewmodel.EventsOverviewUiState
+
+fun getIconFromViewListBool(viewList: Boolean): ImageVector {
+  if (viewList) return Icons.Default.Place else return Icons.AutoMirrored.Filled.List
```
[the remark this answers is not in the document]
nikhen-s: Since it is both used in Hosting and Home, I thought that putting this helper function helps with the readibility. Additionally, if we wanted to change the Icons/were dissatisfied, we just need to change it in this one function and it will propagate.

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -0,0 +1,165 @@
...
+      EventList(
+          uiState.eventList.filteredEvents,
+          Modifier.testTag("eventList").fillMaxWidth().constrainAs(eventList) {
+            top.linkTo(divider.bottom, margin = 8.dp)
+            start.linkTo(parent.start)
+            end.linkTo(parent.end)
+          })
+    } else {
+      EventMap(
+          uiState.eventList.allEvents,
```
jgeaso1266, line 87: nit: `filteredEvents` would also be here

jgeaso1266 (commented):
Go through the SonarCloud one more time and fix any "medium" maintainability issues. But after that the code should be good to go.

app/src/main/java/com/github/se/eventradar/ui/navigation/NavigationActions.kt
```diff
@@ -41,6 +42,16 @@ class NavigationActions(val navController: NavController) {
...
 
+fun getTopLevelDestination(targetRoute: String): TopLevelDestination {
+  val foundDestination = TOP_LEVEL_DESTINATIONS.find { it.route == targetRoute }
+  if (foundDestination != null) {
+    return foundDestination
+  } else {
+    Log.d("Null Error", "Top level destination does not exist for R String Id: $targetRoute")
+    return TOP_LEVEL_DESTINATIONS[2]
+  }
+}
```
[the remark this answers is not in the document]
nikhen-s: Thank you, will remove all code smells from now on before asking for a PR Review.

jgeaso1266 (approved):
LGTM

Nath-Assa (approved):

PR #131, opened 2024-04-18
Updates to the `EventsOverviewViewModel`

Nath-Assa:
### Key Changes
 **Updated EventsOverviewViewModel**:
   - Revised EventsOverviewViewModel to utilize the updated repository interfaces. This adjustment ensures that the ViewModel interacts seamlessly with the underlying data layers, facilitating a more efficient and cleaner data flow within the application.


app/src/main/java/com/github/se/eventradar/model/EventsOverviewViewModel.kt
```diff
@@ -60,27 +60,49 @@ class EventsOverviewViewModel(db: FirebaseFirestore = Firebase.firestore) : View
...
-  private fun getEventContact(contactId: String): String {
-    var contactEmail = ""
-
-    userRef
-        .document(contactId)
-        .collection("private")
-        .limit(1) // Limit the query to retrieve only one document
+  fun getUpcomingEvents(uid: String) {
+    eventRef
+        .whereArrayContains("attendees_list", uid)
```
jgeaso1266, line 65: In theory we keep track of this for each user: https://github.com/swent-sp-2024-event-radar/event-radar-app?tab=readme-ov-file#users-database

jgeaso1266 (commented):
Maybe you can add the User database to your PR, following what's in the readme.

Nath-Assa:
> Maybe you can add the User database to your PR, following what's in the readme.

done

app/src/main/java/com/github/se/eventradar/model/EventsOverviewViewModel.kt
Nath-Assa: > In theory we keep track of this for each user: https://github.com/swent-sp-2024-event-radar/event-radar-app?tab=readme-ov-file#users-database

Great catch! 

> In theory we keep track of this for each user: https://github.com/swent-sp-2024-event-radar/event-radar-app?tab=readme-ov-file#users-database

Great catch! In my latest commit, I optimized getUpcomingEvents to utilize User's eventsAttendeeList, improving query efficiency.

Nath-Assa: > In theory we keep track of this for each user: https://github.com/swent-sp-2024-event-radar/event-radar-app?tab=readme-ov-file#users-database

Great catch!  In my latest commit, I optimized getUpcomingEvents to utilize User's eventsAttendeeList, improving query efficiency.

Nath-Assa:
Note: Tests for the EventOverviewViewModel will be added in the next sprint.

app/src/main/java/com/github/se/eventradar/model/EventsOverviewViewModel.kt
```diff
@@ -60,36 +38,102 @@ class EventsOverviewViewModel(db: FirebaseFirestore = Firebase.firestore) : View
...
-
-    return contactEmail
   }
 
   companion object {
-    fun getLocalDateTime(dateTime: String): LocalDateTime {
+    fun eventFromDocument(document: DocumentSnapshot): Event? {
+      val data = document.data ?: return null
+      return Event(
+          eventName = data["name"] as? String ?: "Unknown Event",
```
jgeaso1266, line 89: nit: I would much rather prefer having some type of error handling instead of default values. If there are default values then we might not catch if there are problems with the database.

app/src/main/java/com/github/se/eventradar/model/EventsOverviewViewModel.kt
```diff
@@ -60,36 +38,102 @@ class EventsOverviewViewModel(db: FirebaseFirestore = Firebase.firestore) : View
...
+          username = data["username"] as? String ?: "Unknown User")
+    }
+
+    private fun getLocalDateTime(dateTime: String): LocalDateTime {
       val formatter = DateTimeFormatter.ofPattern("dd/MM/yyyy HH:mm:ss")
       return LocalDateTime.parse(dateTime, formatter)
     }
 
-    fun getSetOfStrings(data: Any?): Set<String> {
+    private fun getSetOfStrings(data: Any?): Set<String> {
```
jgeaso1266, line 136: nit: this name is very cryptic. What set of strings is this getting? Where is this used?

app/src/main/java/com/github/se/eventradar/model/User.kt
```diff
@@ -0,0 +1,16 @@
+package com.github.se.eventradar.model
+
+data class User(
```
jgeaso1266, line 3: question: was this checked with @christinehahaa?

jgeaso1266 (approved):
some nits and questions.

app/src/main/java/com/github/se/eventradar/model/User.kt
Nath-Assa: yes!

app/src/main/java/com/github/se/eventradar/model/EventsOverviewViewModel.kt
Nath-Assa: I didn't touch this function, but it is useful when you have a field that could either be a single string or a list of strings, and you want to normalize the output to always be a set of strings for consistency. I'll rename it "convertToStringSet".

app/src/main/java/com/github/se/eventradar/model/User.kt
RaulinN: Out of curiosity, could you explain what are the fields 
- eventsAttendeeList & eventsHostList : are you storing the events in which the user is part of / is a host?
- qrCodeUrl (what is the qr code used for exactly?)

RaulinN (commented):
Did not thoroughly review this PR as there's a lot of stuff related to db which will be changed once you have a working implementation of the EventDatabase / UserDatabase

app/src/main/java/com/github/se/eventradar/model/User.kt
Nath-Assa: Yes, the eventsAttendeeList and eventsHostList fields store the IDs of events that the user is attending or hosting, respectively. 
The qrCodeUrl is linked to a unique QR code for each user, which serves multiple purposes: it acts as a digital ticket for events—allowing event staff to scan it to verify the user's ticket purchase—and it can also be used to quickly add the user as a friend by scanning it with a mobile device.

PR #139, opened 2024-04-19
Implement Dependency Injection

jgeaso1266:
## Objective
- separate the data layer from the UI layer
- make our lives easier when testing

## Key Changes
- [ ] add Hilt dependency
- [ ] create database modules for the `Users` and `Events` databases
- [ ] create repository interfaces for both databases
- [ ] create mock and Firebase class implementations (though not all are implemented - to be done in separate PRs)
- [ ] implemented the `FirebaseEventRepository` as an example


app/src/main/java/com/github/se/eventradar/model/data/EventDatabase.kt
```diff
@@ -0,0 +1,18 @@
...
+import dagger.hilt.InstallIn
+import dagger.hilt.components.SingletonComponent
+
+@Module
+@InstallIn(SingletonComponent::class)
+abstract class EventDatabase {
+  @Binds
+  abstract fun bindFirebaseEventRepository(userRepository: FirebaseEventRepository): EventRepository
+
+  @Binds abstract fun bindMockUserRepository(userRepository: MockEventRepository): EventRepository
```
nikhen-s, line 17: imp: I think there was a typo for this and maybe renaming the parameter to something other than userRepository could help avoid confusion, something like bindMockEventRepository(mockEventRepository: MockEventRepository)

nikhen-s (changes requested):
Looks great overall, just 2 important comments in `EventDatabase.kt` and `Resource.kt` + some nits with naming convention.

app/src/main/java/com/github/se/eventradar/model/repository/FirebaseEventRepository.kt
```diff
@@ -0,0 +1,132 @@
...
+import com.github.se.eventradar.model.getLocation
+import com.google.firebase.firestore.CollectionReference
+import com.google.firebase.firestore.FirebaseFirestore
+import com.google.firebase.firestore.ktx.firestore
+import com.google.firebase.ktx.Firebase
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
+import kotlinx.coroutines.tasks.await
+
+class FirebaseEventRepository : EventRepository {
```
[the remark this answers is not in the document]
nikhen-s: imp: I think we should add @Inject constructor here so that Hilt knows how to provide Instances of FirebaseEventRepository (same with the FIrebaseUserRepository, MockEventRepository and MockUserRepository)

app/build.gradle.kts
```diff
@@ -207,3 +210,7 @@ sonar {
         property("sonar.coverage.jacoco.xmlReportPaths", "build/reports/jacoco/jacocoTestReport/jacocoTestReport.xml")
     }
 }
+
+kapt {
+    correctErrorTypes = true
+}
```
nikhen-s, line 216: nice! didn't know we could do this :)

app/src/main/java/com/github/se/eventradar/model/Resource.kt
```diff
@@ -0,0 +1,7 @@
+package com.github.se.eventradar.model
+
+sealed class Resource<out T> {
+  data class Success<out T>(val data: T) : Resource<T>()
+
```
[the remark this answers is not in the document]
jgeaso1266: It could be beneficial but might be overkill for our purposes.

app/src/main/java/com/github/se/eventradar/model/repository/FirebaseEventRepository.kt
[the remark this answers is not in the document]
jgeaso1266: > nit: maybe we could rename this to FIrebaseEventRepositoryImpl too for redability? (Apologies for the many nits on naming)

In my opinion, adding Impl doesn't make it any more readable. Adding a descriptive word or two at the beginning (like `Firebase` or `Mock`) goes a much farther way in helping the developer understand the purpose of the class.

app/src/test/java/com/github/se/eventradar/MockEventRepositoryUnitTest.kt
```diff
@@ -0,0 +1,132 @@
...
+import com.github.se.eventradar.model.event.EventTicket
+import com.github.se.eventradar.model.repository.event.IEventRepository
+import com.github.se.eventradar.model.repository.event.MockEventRepository
+import java.time.LocalDateTime
+import kotlinx.coroutines.test.runTest
+import org.junit.Before
+import org.junit.Test
+
+class MockEventRepositoryUnitTest {
+  private lateinit var eventRepository: IEventRepository
```
RaulinN, line 16: Ok that's nice! Two things to keep in mind:
- be careful with concurrency => tests will execute in parallel
- with this approach, you are usnig **manual dependency injection** (i.e. you are saying "I create a repository of type IEventRepository" and test it). In other words, you are using Hilt (automated injection) for the app, and manual injection for the tests. Since you defined the `@Binds` in the modules above, you should be able to write the following

```kotlin
@HiltAndroidTest
@RunWith(AndroidJUnit4::class)
class MockEventRepositoryUnitTest {

    @get:Rule
    var hiltRule = HiltAndroidRule(this)

    @Before
    fun setUp() {
        hiltRule.inject() // inject dependencies using the module for your tests
    }

    // write tests below
}
```

app/src/main/java/com/github/se/eventradar/model/data/EventDatabaseModule.kt
```diff
@@ -0,0 +1,20 @@
...
+@Module
+@InstallIn(SingletonComponent::class)
+abstract class EventDatabaseModule {
+  @Binds
+  abstract fun bindFirebaseEventRepository(
+      userRepository: FirebaseEventRepository
+  ): IEventRepository
+
+  @Binds abstract fun bindMockEventRepository(userRepository: MockEventRepository): IEventRepository
+}
```
RaulinN, line 20: Ok and here are a few things regarding the hilt modules (both event and user)
- first of all, I really like the method using `@Binds`. Some people differentiate between tests/app using `@Named` and it's quite ugly / impractical imo
- they are typically placed in a folder called `di` (for dependency injection)
- I am not the best with Hilt, but I don't think this will work (test it before, I might be wrong because I rarely worked with `@Binds`). To my understanding, you should have:
  - a module for the events in the `di` folder (e.g. `EventDatabaseModule`) with a `@Binds` and a `@Provides` (without the `@Provides`, Hilt will not know how to create an `IEventRepository` in this case)
  - a module for the events in the tests specifically (e.g. `EventDatabaseTestModule`) with a `@Binds` and a `@Provides`

> [!TIP]
> You can find an example [here](https://github.com/RaulinN/swent-snippets/blob/db561acbcbb3841b791343387e1441eb6229a69e/what-is-dependency-injection.kt#L73) (starting at line 73)

app/src/main/java/com/github/se/eventradar/model/repository/event/IEventRepository.kt
```diff
@@ -0,0 +1,16 @@
...
+interface IEventRepository {
+  suspend fun getEvents(): Resource<List<Event>>
+
+  suspend fun getEvent(id: String): Resource<Event?>
+
+  suspend fun addEvent(event: Event): Resource<Boolean>
+
+  suspend fun updateEvent(event: Event): Resource<Boolean>
+
+  suspend fun deleteEvent(event: Event): Resource<Boolean>
```
RaulinN, line 15: Depending on how you intend to use this interface, you might not need to return `Boolean`, but `Unit` instead. For instance (e.g. `addEvent`), you could have
- the db could add the event => Resource.Success(Unit)
- the db could not add the event => Resource.Failure(error)

RaulinN: _(this is kinda nitpicking)_

app/src/main/java/com/github/se/eventradar/model/repository/event/MockEventRepository.kt
```diff
@@ -0,0 +1,63 @@
...
+              ticket = EventTicket("Test Ticket", 0.0, 1),
+              contact = "Test Contact Email",
+              organiserList = setOf("Test Organiser"),
+              attendeeList = setOf("Test Attendee"),
+              category = EventCategory.COMMUNITY,
+              fireBaseID = "$it")
+        }
+
+    return Resource.Success(events)
+  }
```
RaulinN, line 31: This is crucial! **This repository will** (ideally) be utilized in the tests (instead of making connections with firestore), even if you're conducting instrumented tests. During these instrumented tests, you'll probably need to add/remove events and verify if the UI responds accurately to these modifications. 

However, with your current implementation of `MockEventRepository`, this won't be feasible. I recommend transforming `MockEventRepository` into an in-memory database, where data is stored in a map. If you end up doing this, make sure you implement the other methods as well.

```suggestion
class MockEventRepository constructor() : IEventRepository {
  private val storage: MutableMap<String, Event> = mutableMapOf()

  override suspend fun getEvents(): Resource<List<Event>> {
    val events = storage.values.toList()
    return Resource.Success(events)
  }
```

> [!TIP]
> Take a look at the first 40 lines of [this file](https://github.com/RaulinN/swent-snippets/blob/migration-hilt-full-file/hilt-migration-from-manual-di.kt)

Note that you can obviously add default events in your map if you wish so

app/src/main/java/com/github/se/eventradar/model/repository/event/FirebaseEventRepository.kt
RaulinN, line 1: That's quite good overall! One small thing: you are converting Events to maps twice (repeated like 30 lines of code). I would put that somewhere else (e.g. have a set of methods in the Event class so that you can do `event.toMap()` and `Event.fromMap(map)`, or have private utility functions in this file if you believe you will not use it anywhere else

app/src/main/java/com/github/se/eventradar/model/Resource.kt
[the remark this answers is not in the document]
RaulinN: > imp: Do you think a Loading data class could still be beneficial to show user feedback when the data is still being fetched in the asynchronous operation?

This will not work since your functions are `suspend` functions

app/src/main/java/com/github/se/eventradar/model/Resource.kt
RaulinN, line 1: Again, really good! This `Resource` class is great! If you want a slightly more robust Resource class, I would recommend wathcing **the first 8 minutes of this video** : https://www.youtube.com/watch?v=MiLN2vs2Oe0

- 00:00 -> 08:00 explanation of the updated Resource class
- 08:00 -> 14:00 example usage

RaulinN (commented):
Hey there! Fantastic work getting started! I've included a few comments that should come in handy. Don't worry about the number of comments —I just wanted to offer some guidance. The code you've written is an excellent foundation to build upon.

Note: You're currently utilizing automated dependency injection (Hilt). However, manual dependency injection is also an option if you prefer. I've just written a few code snippets on Discord that might be helpful if you want to see the difference in implementation.

app/src/main/java/com/github/se/eventradar/model/data/EventDatabaseModule.kt
jgeaso1266: Yes I think what I have won't work. I believe I can accomplish this just using `@Provides`.

app/src/test/java/com/github/se/eventradar/MockEventRepositoryUnitTest.kt
jgeaso1266: I don't believe I am able to do this in a unit test (under `test`, not `androidTest`. Since I am only explicitly testing the mock database, I think it is fine to not use the injection logic here.

nikhen-s (approved):
LGTM. Great changes and I think we have the relevant interfaces now to continue implementing tests for the `Event` and `User` fire store databases.

PR #149, opened 2024-04-23
Qr code scanner

wisdomseeker2001:
Hi,

I have added a QRCode Screen Ui, it now has what the final version should look like.

I have not implemented the screen for MyQRCode which Is part of it upon selecting another tab (waiting on Achille, would be happy to be assigned to this task following sprint).

The Scanning QR Code tab (in QrCodeScreen.kt) that I implemented will be used in both: Scanning Tickets and Scanning new Friends:
-For adding a friend, once QR code scanned, should directly open screen to send them a message. 
-For Scanning Ticket, there should be a text saying 'ENTRY or 'NO ENTRY' that pops up
the above two points haven't been coded because apparently this is a separate task that should be done when doing the backend logic of scanning a qR code

qrCodeScan.Kt is basically the constraints and instructions of how the QR Code should interpret the image. 
It takes takes a callback function which will be passed by the caller (I'd be more than happy to be assigned to these tasks for the next sprints - backend of ticket scanned and of new friend scanned/added)

I also took care of asking for camera permissions but I think this needs to be worked on a little more and handle the cases where the user refuses access... let me know your thoughts on this. 

One thing I was not able to do is space the camera view evenly between the navigation and tab bar - I tried many things, but it seems to stay where it is, still looks good in my opinion

Joyti


app/src/androidTest/java/com/github/se/eventradar/qrCode/qrCodeScanFriendUiTest.kt
achillet3, line 1: Good tests for the UI and the use of the camera! However, you are missing a little coverage for the QR Code Scanner... There is very little coverage of QR Code Constraints. It is hard to test perfectly but you could feed it manually a QR Code of something you are sure of, for example, the QR Code from the "QR Codes" folder in Firebase storage (AwOXI3dCWjfYKk7bnBQ0S94WxbD2-20240427045836.png) encodes: 
- uid = AwOXI3dCWjfYKk7bnBQ0S94WxbD2
- date = 20240427045836

app/src/main/java/com/github/se/eventradar/qrCode/QrCodeScanConstraints.kt
achillet3, line 1: Little to no coverage...

app/src/main/java/com/github/se/eventradar/ui/qrCode/QrCodeScanCamera.kt
achillet3, line 1: LGTM! Well done for the implementation and the testing of the camera

app/src/main/java/com/github/se/eventradar/ui/qrCode/qrCodeScanFriendUi.kt
```diff
@@ -0,0 +1,144 @@
...
+private val dummyQrCodeScanned: (String) -> Unit = { qrCode ->
+  Log.d("QRCodeScanner", "QR Code Scanned: $qrCode")
+  // You can perform any additional logic here for testing
+}
+
+@androidx.compose.ui.tooling.preview.Preview
+@Composable
+fun QrcodeScanTest() {
+  QrCodeScreen(NavigationActions(rememberNavController()))
+}
```
achillet3, line 144: You could comment out this portion of the code because it is unnecessary in the main branch. It will help with test coverage as well

app/src/main/java/com/github/se/eventradar/ui/qrCode/qrCodeScanFriendUi.kt
```diff
@@ -0,0 +1,144 @@
...
+                  top.linkTo(logo.bottom, margin = 16.dp)
+                  start.linkTo(parent.start)
+                  end.linkTo(parent.end)
+                }
+                .testTag("tabs"),
+        contentColor = MaterialTheme.colorScheme.primary) {
+          Tab(
+              selected = selectedTabIndex == 0,
+              onClick = { selectedTabIndex = 0 },
+              modifier = Modifier.testTag("My QR Code"),
```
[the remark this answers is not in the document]
jgeaso1266: Test tags and names are different, will be a little more difficult to refactor out.

app/src/main/java/com/github/se/eventradar/ui/qrCode/qrCodeScanFriendUi.kt
jgeaso1266: This in theory should not be included in coverage, but I see no problem with keep it in the main branch. A future dev will inevitably want to do something like this.

PR #150, opened 2024-04-23
Quick event null fix

achillet3:
A quick fix to EventsOverviewViewModel. This is solely to get the app up and running fixing a bad PR from the last sprint. Deleted incorrectly formatted event documents in Firebase as well.


wisdomseeker2001 (approved):
Seems all good to me, anyways this class needs to change completely to adapt to the code's new structure

PR #155, opened 2024-04-25
Event overview view model final version

Nath-Assa:
### Key changes
-  Reinstated the previously implemented getEvents and getUpcomingEvents methods(from #131) in EventsOverviewViewModel, which were removed in a prior merge. The ViewModel utilizes the revised repository interfaces. 
- Added tests for EventOverviewViewModel
- Updated home for DI
- Fixed typo in EventList

### DIFF
--- a/app/src/main/java/com/github/se/eventradar/model/EventsOverviewViewModel.kt
+++ b/app/src/main/java/com/github/se/eventradar/model/EventsOverviewViewModel.kt
@@ -19,58 +19,58 @@ constructor(
     private val eventRepository: IEventRepository,
     private val userRepository: IUserRepository
 ) : ViewModel() {
-    private val _uiState = MutableStateFlow(EventsOverviewUiState())
-    val uiState: StateFlow<EventsOverviewUiState> = _uiState
+  private val _uiState = MutableStateFlow(EventsOverviewUiState())
+  val uiState: StateFlow<EventsOverviewUiState> = _uiState
 
-    fun getEvents() {
-        viewModelScope.launch {
-            when (val response = eventRepository.getEvents()) {
-                is Resource.Success -> {
-                    _uiState.value =
-                        _uiState.value.copy(
-                            eventList =
+  fun getEvents() {
+    viewModelScope.launch {
+      when (val response = eventRepository.getEvents()) {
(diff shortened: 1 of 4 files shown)


Nath-Assa:
All tests pass when ran individually but testGetEventsSuccess() fails when I run all of them. Will look into it.

app/src/main/java/com/github/se/eventradar/model/EventsOverviewViewModel.kt
```diff
@@ -1,21 +1,76 @@
...
+      }
+    }
+  }
 
-  fun getEvents() {}
+  fun getUpcomingEvents(uid: String) {
+    viewModelScope.launch {
+      when (val userResponse = userRepository.getUser(uid)) {
+        is Resource.Success -> {
+          val user = userResponse.data ?: return@launch
```
jgeaso1266, line 44: I'm not familiar with the syntax, what does it do?

app/src/test/java/com/github/se/eventradar/EventOverviewViewModelTest.kt
```diff
@@ -0,0 +1,189 @@
...
+    eventRepository = MockEventRepository()
+    userRepository = MockUserRepository()
+    viewModel = EventsOverviewViewModel(eventRepository, userRepository)
+  }
+
+  @Test
+  fun testGetEventsEmpty() = runTest {
+    viewModel.getEvents()
+
+    advanceUntilIdle()
```
jgeaso1266, line 94: nit: I don't believe this is necessary. This is handled automatically by `runTest`.

jgeaso1266 (approved):
One nit and one question, but otherwise LGTM.

app/src/main/java/com/github/se/eventradar/model/EventsOverviewViewModel.kt
Nath-Assa: If the user is null,  return@launch instructs the function to return immediately from the launch coroutine scope. This prevents any subsequent operations that would rely on a non-null user object.
After further review, there is no need for ?: return@launch since data is guaranteed non-null on success of the getUser(uid) function.

PR #170, opened 2024-04-26
Update user.kt to have Set of Friends

wisdomseeker2001:
added Set of Friends
changed List of Events user is attending or hosting to set as their should be no duplicates

### DIFF
--- a/app/src/main/java/com/github/se/eventradar/model/User.kt
+++ b/app/src/main/java/com/github/se/eventradar/model/User.kt
@@ -8,8 +8,9 @@ data class User(
     val lastName: String,
     val phoneNumber: String,
     val accountStatus: String,
-    val eventsAttendeeList: List<String>,
-    val eventsHostList: List<String>,
+    val eventsAttendeeList: Set<String>,
+    val eventsHostList: Set<String>,
+    val friendList: Set<String>,
     val profilePicUrl: String,
     val qrCodeUrl: String,
     val username: String,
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/github/se/eventradar/model/User.kt
```diff
@@ -49,10 +51,10 @@ data class User(
   }
 }
 
-private fun convertToListOfStrings(data: Any?): List<String> {
+private fun convertToListOfStrings(data: Any?): MutableSet<String> {
```
Nath-Assa, line 54: Rename function to: "convertToMutableSetOfStrings"

app/src/main/java/com/github/se/eventradar/model/User.kt
```diff
@@ -8,8 +8,9 @@ data class User(
     val lastName: String,
     val phoneNumber: String,
     val accountStatus: String,
-    val eventsAttendeeList: List<String>,
-    val eventsHostList: List<String>,
+    val eventsAttendeeList: MutableSet<String>,
```
Nath-Assa, line 11: To maintain consistency across the application, if you think using MutableSet is the better option here, then we should also consider changing the organiserList and attendeeList properties in the Event class from Set to MutableSet. This adjustment would align these properties with the same mutable approach.

app/src/main/java/com/github/se/eventradar/model/User.kt
```diff
@@ -49,10 +51,10 @@ data class User(
   }
```
Nath-Assa, line 51: toMap() function does not map the new "friendList" property.

Nath-Assa (changes requested):
I've added a few comments and suggested edits for your consideration. Please ensure all checks in the continuous integration (CI) process pass before merging.

app/src/main/java/com/github/se/eventradar/model/User.kt
wisdomseeker2001: thanks you for all the comments, will make the changes now

Nath-Assa:
Resolved in #177

PR #173, opened 2024-04-28
Hosted Events ViewModel + Tests

### LINKED ISSUE #109: Create "My Hosting": ViewModel
## Link To User Story and Figma
User Story #76, Figma Issue #98 

## Objective
- Implement a ViewModel that responds to changes in Database when user refreshes the "My Hosting" page.

## Acceptance Criteria
- [ ] Connect the backend and the frontend through a ViewModel. (Will refine)

nikhen-s:
Related to #109 => closes #109 
## Objective
- Create a ViewModel for User's Hosted Events

## Changes made
- Created a new package for viewmodels for better separation of classes
- Created HostedEventsViewModel and HostedEventsViewModelTest to test the functionality
- Added Enum file for Map view/List View and Upcoming/Browse Events Tab to be used in the viewModel
- ViewModel now handles all state changes in Home.kt UI and Hosting.kt UI

Notes:
- This is a commit just for the ViewModel, thus all the UI changes for Hosting.kt in order to utilize the new ViewModel which keeps the Map/List View Type and the Upcoming/Browse Tab will be made in the HostedEventsUI PR After this.

### DIFF
--- a/app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
+++ b/app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
@@ -44,7 +44,7 @@ constructor(
         is Resource.Success -> {
           val user = userResponse.data!!
           val attendeeList = user.eventsAttendeeList
-          if (attendeeList.isNotEmpty()) {
+          if (attendeeList.isNotEmpty()) { //will it ever be that
             when (val events = eventRepository.getEventsByIds(attendeeList)) {
               is Resource.Success -> {
                 _uiState.value =
@@ -76,7 +76,7 @@ constructor(
 
 data class EventsOverviewUiState(
     val eventList: EventList = EventList(emptyList(), emptyList(), null),
-    var viewType: ViewType = ViewType.LIST,
+    var viewList: Boolean = true,
     var tab: Tab = Tab.BROWSE,
     val searchQuery: String = "",
 )
--- a/app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
+++ b/app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
@@ -76,5 +76,12 @@ constructor(
 
 data class EventsOverviewUiState(
     val eventList: EventList = EventList(emptyList(), emptyList(), null),
+    var viewType: ViewType = ViewType.LIST,
+    var tab: Tab = Tab.BROWSE,
     val searchQuery: String = "",
 )
+
+enum class Tab {
+    BROWSE,
+    UPCOMING_EVENTS
+}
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/github/se/eventradar/viewmodel/HostedEventsViewModel.kt
```diff
@@ -0,0 +1,66 @@
...
+                                events.data, events.data, _uiState.value.eventList.selectedEvent))
+              }
+              is Resource.Failure -> {
+                Log.d("HostedEventsViewModel", "Error getting hosted events for $uid")
+                _uiState.value =
+                    _uiState.value.copy(eventList = EventList(emptyList(), emptyList(), null))
+              }
+            }
+          } else {
+            _uiState.value =
```
jgeaso1266, line 48: Is this necessary? I think the `eventList` variable is already initialized with these values.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
```diff
@@ -75,5 +76,12 @@ constructor(
 
 data class EventsOverviewUiState(
     val eventList: EventList = EventList(emptyList(), emptyList(), null),
+    var viewType: ViewType = ViewType.LIST,
+    var tab: Tab = Tab.BROWSE,
     val searchQuery: String = "",
 )
+
+enum class Tab {
```
jgeaso1266, line 84: Good idea! Might be overkill for just two tabs but is a good way to make it scalable to more tabs.

app/src/main/java/com/github/se/eventradar/viewmodel/ReusableEnums.kt
```diff
@@ -0,0 +1,6 @@
+package com.github.se.eventradar.viewmodel
+
+enum class ViewType {
```
jgeaso1266, line 3: Could this be accomplished with just one boolean variable?

jgeaso1266 (commented):
A few comments but looks good overall. Not approving yet because I have some questions I would like you to answer.

app/src/main/java/com/github/se/eventradar/viewmodel/ReusableEnums.kt
nikhen-s: Yes it can, i think that would make more sense too. Will change.

app/src/main/java/com/github/se/eventradar/viewmodel/HostedEventsViewModel.kt
nikhen-s: I think that it is needed because in the situation where we delete a hosted event from the database, then we call getHostedEvents(uid), if it previously had a non-empty EventList value, it could be recomposed with this even though the event is no longer in the database as it takes the previous EventList value (which will not be updated without this else part), because the empty eventList is only initialized. I'm assuming this because the viewModel state is kept across recompositions (but I'm not sure if this is right)

jgeaso1266 (approved):
LGTM

PR #180, opened 2024-04-30
Exclude Previews from Code Coverage

jgeaso1266:
## Objectives
- exclude preview functions from code coverage

## Key Changes
- defined a new annotation to annotate preview files (see [here](https://stackoverflow.com/questions/47824761/how-would-i-add-an-annotation-to-exclude-a-method-from-a-jacoco-code-coverage-re))
- annotate all preview files currently in main with this annotation

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -186,7 +186,6 @@ tasks.register("jacocoTestReport", JacocoReport::class) {
         "**/*Test*.*",
         "android/**/*.*",
         "**/SignatureChecks.*",
-        "**/*Preview*.*",
     )
     val debugTree = fileTree("${project.buildDir}/tmp/kotlin-classes/debug") {
         exclude(fileFilter)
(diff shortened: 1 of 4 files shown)


app/build.gradle.kts
```diff
@@ -186,7 +186,6 @@ tasks.register("jacocoTestReport", JacocoReport::class) {
         "**/*Test*.*",
         "android/**/*.*",
         "**/SignatureChecks.*",
-        "**/*Preview*.*",
```
nikhen-s, line 189: Is there a particular reason why this was removed? I thought that it could still help filter out the `"**/*Preview*.*"` files.

nikhen-s (approved):
Just 1 nit question, but otherwise LGTM.

achillet3 (approved):
LGTM

app/build.gradle.kts
jgeaso1266: I took it out because there are no Preview files. I originally added this because I thought it would remove preview _functions_ from coverage, but that was not the case.

PR #181, opened 2024-04-30
QR CODE MVVM MODEL (UPDATED)  UI + VM

wisdomseeker2001:
Last PR#149 has now become redundant given I had to redesign the whole code. 
In This PR, 4 files are to be of concern, namely:

Note there is a **Weird Error: Even though I am up to date with the main, home.kt has an error and says hilt is not being recognised :/**

1) QrCodeFriendUi.kt (UI for Screen of QrCode App Section):  Implemented UI for Scanning Qr Screen 
2) QrCodeCamera.kt (UI of camera display itself): Separated it so that Scanning QR for events can also call this it.
3) QrCodeFriendViewModel
4) QrCodeAnalyser 

I also put in 4) in the ViewModel package because it is technically taking care of backend logic and working along side the ViewModel... please look at explanation below that explains the implementation and let me know if you have a better location to suggest:

_QrCodeCamera needs to pass Analyser as argument to its android view to tell the image processing how to analyse. The Analyser must override the Analyse() function which defines the analysing constraints and must call a callback on the decoded result  encoded in QR Code.

Given it's the Viewmodel that should communicate directly with the UI, I did the following: ViewModels init block: sets 
the callback 'onDecoded' in the Analyser class to a function that initialises the MutableStateFlow<String?> '_decodedResult' in the Viewmodel to the  string encoded in the QR Code (the Friend's UID). 

Given Viewmodel is initialised before UI rendered, we are sure that by the time the QR is analysed, the callback will be set correctly and hence the Viewmodel will have access to the encoded string via its private val _decodedResult.

The Viewmodel only calls its function updateFriendList (communicates with Firebase Repository) **iff** _decodedResult is non null (only once QR code has been decoded). Once updatedFriendList has been executed and has successfully updated the Firebase, it will set NavigationEvent to NavigateToNextScreen which will signal the UI to call the navigation and switch screen to send a private message to the new Friend. 

This was not so intuitive at first and took me time to find the cleanest way to implement it. 
ViewModel & UI can be improved on by having a state where the UI reflects a loading icon if firebase operations take a long time
Would also need to handle the case where the updates really do fail within 1 minute's time. Personally think this should be an extra once its current functionality works

Any Feedback for this rather complex PR will be greatly appreciated as still a huge amount of testing to be done and Hilt to be implemented


app/src/main/java/com/github/se/eventradar/ui/qrCode/QrCodeCamera.kt
```diff
@@ -0,0 +1,90 @@
...
+                        preview,
+                        imageAnalysis)
+              } catch (e: Exception) {
+                e.printStackTrace()
+              }
+              previewView
+            },
+            modifier = Modifier.weight(1.5f).aspectRatio(1f).padding(horizontal = 32.dp)
+            //                Text(
+            //                    text = code, )
```
jgeaso1266, line 85: nit: remove these lines of code

app/src/main/java/com/github/se/eventradar/ui/qrCode/Qr code scanner by wisdomseeker2001 · Pull Request #149 · swent-sp-2024-event-radar:event-radar-app.html
jgeaso1266, line 1: This must be deleted!

app/src/main/java/com/github/se/eventradar/ui/qrCode/QrCodeCamera.kt
```diff
@@ -0,0 +1,90 @@
...
+              try {
+                cameraFutureProvider
+                    .get()
+                    .bindToLifecycle(
+                        lifeCycleOwner, // only launched during current compose lifecycle
+                        selector,
+                        preview,
+                        imageAnalysis)
+              } catch (e: Exception) {
+                e.printStackTrace()
```
jgeaso1266, line 79: It would be better to log a message (using `Log.e` or `Log.d`)

app/src/main/java/com/github/se/eventradar/ui/qrCode/QrCodeFriendUi.kt
```diff
@@ -0,0 +1,165 @@
...
+import com.github.se.eventradar.ui.navigation.TopLevelDestination
+import com.github.se.eventradar.viewmodel.qrCode.NavigationEvent
+import com.github.se.eventradar.viewmodel.qrCode.QrCodeFriendViewModel
+
+@Composable
+fun QrCodeScreen(viewModel: QrCodeFriendViewModel, navigationActions: NavigationActions) {
+  val navigateState by viewModel.navigationEvent.collectAsState()
+
+  // React to changes in navigation state
+  LaunchedEffect(navigateState) {
```
jgeaso1266, line 49: This is not necessary. You can simply navigate right where needed in the logic below using `navigationActuions.navigate({route})`

app/src/main/java/com/github/se/eventradar/ui/qrCode/QrCodeFriendUi.kt
```diff
@@ -0,0 +1,165 @@
...
+                textId = R.string.message_chats,
+            )) // Adjust according to your actual navigation logic
+        viewModel.resetNavigationEvent() // Reset the navigation event in the ViewModel to prevent
+        // repeated navigations
+      }
+      else -> Unit // Do nothing if the state is None or any other non-navigational state
+    }
+  }
+
+  var selectedTabIndex by remember { mutableIntStateOf(0) }
```
jgeaso1266, line 65: This can be in the uiState and edited by the view model. (I believe @nikhen-s has an example of this in his Hosting screen code)

app/src/main/java/com/github/se/eventradar/ui/qrCode/QrCodeCamera.kt
```diff
@@ -0,0 +1,90 @@
...
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.platform.LocalLifecycleOwner
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.viewinterop.AndroidView
+import androidx.core.content.ContextCompat
+import com.github.se.eventradar.viewmodel.qrCode.QrCodeAnalyser
+
+class QrCodeCamera {
```
jgeaso1266, line 31: What is the point of this class? There are no variables and no constructor, so is this needed?

app/src/main/java/com/github/se/eventradar/ui/qrCode/QrCodeFriendUi.kt
```diff
@@ -0,0 +1,165 @@
...
+            })
+  }
+}
+
+// private val dummyQrCodeScanned: (String) -> Unit = { qrCode ->
+//  Log.d("QRCodeScanner", "QR Code Scanned: $qrCode")
+//  // You can perform any additional logic here for testing
+// }
+
+// @androidx.compose.ui.tooling.preview.Preview
```
jgeaso1266, line 161: You can leave this uncommented and add the annotation described in #180 (once it's approved)

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/QrCodeAnalyser.kt
```diff
@@ -0,0 +1,59 @@
...
+            MultiFormatReader()
+                .apply {
+                  setHints(
+                      mapOf(DecodeHintType.POSSIBLE_FORMATS to arrayListOf(BarcodeFormat.QR_CODE)))
+                }
+                .decode(binaryBitmap)
+        // if onDecoded is null (has not been initialised b Viewmodel) will simply return Null
+        onDecoded?.invoke(result.toString())
+      } catch (e: Exception) {
+        e.printStackTrace()
```
jgeaso1266, line 47: Again, better to use `Log` here.

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/QrCodeFriendViewModel.kt
```diff
@@ -0,0 +1,116 @@
...
+import kotlinx.coroutines.async
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.filterNotNull
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withTimeoutOrNull
+
+enum class NavigationEvent {
```
jgeaso1266, line 20: This isn't necessary

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/QrCodeFriendViewModel.kt
```diff
@@ -0,0 +1,116 @@
...
+// firebase operations take a long time
+// would also need to handle the case where the updates really do fail within 1 minute's time.
+// personally gonna make this extra.
+
+@HiltViewModel
+class QrCodeFriendViewModel
+@Inject
+constructor(
+    private val firebaseUserRepository: IUserRepository, // Dependency injection
+    qrCodeAnalyser: QrCodeAnalyser // Dependency injection
```
jgeaso1266, line 34: Does this need to be in the constructor? I don't see a mock implementation of this.

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/QrCodeFriendViewModel.kt
```diff
@@ -0,0 +1,116 @@
...
+            }
+          }
+    }
+  }
+
+  private fun updateFriendList(friendID: String) {
+    viewModelScope.launch {
+      val myUID = FirebaseAuth.getInstance().currentUser!!.uid
+
+      val friendUserDeferred = async { firebaseUserRepository.getUser(friendID) }
```
jgeaso1266, line 62: I believe this could all go into one `runBlocking` block (but needs to be tested)

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/QrCodeAnalyser.kt
```diff
@@ -0,0 +1,59 @@
...
+      // parameters to scan
+      val source =
+          PlanarYUVLuminanceSource(
+              bytes, image.width, image.height, 0, 0, image.width, image.height, false)
+
+      val binaryBitmap = BinaryBitmap(HybridBinarizer(source))
+      try {
+
+        // result is the info encoded in to QR Code (String of userId in our case)
+        val result =
```
jgeaso1266, line 37: One thing to think about as an extra (or possibly to write in PRD) is the security of this process, and its proneness to hacks and other attacks.

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/QrCodeFriendViewModel.kt
```diff
@@ -0,0 +1,116 @@
...
+  val navigationEvent: StateFlow<NavigationEvent> = _navigationEvent.asStateFlow()
+
+  init {
+    qrCodeAnalyser.onDecoded = { decodedString ->
+      _decodedResult.value = decodedString ?: "Failed to decode QR Code"
+    }
+
+    viewModelScope.launch {
+      _decodedResult
+          .filterNotNull() // Only continue if the value is not null
```
jgeaso1266, line 49: We also do not want it to continue if the value is "Failed to decode QR Code"

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/QrCodeFriendViewModel.kt
```diff
@@ -0,0 +1,116 @@
...
+// ViewModel & UI can be improved on by having a state where the UI reflects a loading icon if
+// firebase operations take a long time
+// would also need to handle the case where the updates really do fail within 1 minute's time.
+// personally gonna make this extra.
+
+@HiltViewModel
+class QrCodeFriendViewModel
+@Inject
+constructor(
+    private val firebaseUserRepository: IUserRepository, // Dependency injection
```
jgeaso1266, line 33: you should simply name this `userRepository` as it could also be the mock version.

build.gradle.kts
```diff
@@ -9,6 +9,15 @@ buildscript {
     val hiltVersion by rootProject.extra { "2.51.1" }
     dependencies {
         classpath("com.google.android.libraries.mapsplatform.secrets-gradle-plugin:secrets-gradle-plugin:2.0.1")
+
+
+        classpath ("com.android.tools.build:gradle:7.0.4")
+        classpath ("org.jetbrains.kotlin:kotlin-gradle-plugin:1.5.31")
+
+        classpath("com.google.dagger:hilt-android-gradle-plugin:$hiltVersion")
```
jgeaso1266, line 17: this is here twice. Make sure there are no duplicates in this file.

app/src/main/java/com/github/se/eventradar/ui/qrCode/QrCodeCamera.kt
```diff
@@ -0,0 +1,90 @@
...
+              val selector =
+                  CameraSelector.Builder()
+                      .requireLensFacing(CameraSelector.LENS_FACING_BACK)
+                      .build()
+              preview.setSurfaceProvider(previewView.surfaceProvider)
+              val imageAnalysis =
+                  ImageAnalysis.Builder()
+                      .setBackpressureStrategy(ImageAnalysis.STRATEGY_KEEP_ONLY_LATEST)
+                      .build()
+              imageAnalysis.setAnalyzer(ContextCompat.getMainExecutor(context), QrCodeAnalyser())
```
jgeaso1266, line 68: This is a different analyser from the one in the view model. Does that matter?

jgeaso1266 (dismissed):
Has this been tested? I'm not familiar with the tools/libraries used here, so I can't be of much help there. But a lot of my comments are for general things in the other parts of the code (non qr related). Also, you should add unit tests for any helper/non-UI functions (in the `test` folder, not `androidTest`), especially those dealing directly with the scanner. Also, I am confused how the `QrCodeAnalyser` and the `QrCodeScanner` work together. Maybe we can talk offline about it.

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/QrCodeAnalyser.kt
wisdomseeker2001: Actually need to probably do a some sort of check: as you said if someone displays a random QR code they generated, that encodes a string... could be an attack.. need to think of some condition / property shared amongst all UIDs, could be length for e.g. if they all have a fixed length .... wouldn't eliminate attacks but defo cut it down a lot! should I do this or keep as extra?

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/QrCodeFriendViewModel.kt
wisdomseeker2001: this is exactly why it is followed by "if (result != "Failed to decode QR Code")"

app/src/main/java/com/github/se/eventradar/ui/qrCode/QrCodeCamera.kt
wisdomseeker2001: it is the same :)

wisdomseeker2001: Very Very Good Catch, thank you for spotting this one out... have done it via Hilt DI now

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/QrCodeAnalyser.kt
```diff
@@ -0,0 +1,60 @@
...
+import java.nio.ByteBuffer
+import javax.inject.Inject
+
+class QrCodeAnalyser @Inject constructor() : ImageAnalysis.Analyzer {
+
+  // list of supported Image Formats
+  private val supportedImageFormats =
+      listOf(ImageFormat.YUV_420_888, ImageFormat.YUV_422_888, ImageFormat.YUV_444_888)
+
+  var onDecoded: ((String?) -> Unit)? = null //
```
Viccha18, line 21: Is this still a TODO ?
edit : either comment "Set by the viewModel" or create a Setter function with `private var onDecoded` for more clarity that this need to be set from the outside by the entity that create an instance of QrCodeAnalyser, or even add it as an argument in the constructor of the class (also not sure why you use @Inject on the constructor of that class)

app/src/test/java/com/github/se/eventradar/FirebaseUserRepositoryUnitTest.kt
```diff
@@ -215,83 +215,84 @@ class FirebaseUserRepositoryUnitTest {
...
-    val message = "Exception"
-    every { userRef.document(uid).set(any()) } returns mockTask(null, Exception(message))
-
-    val result = firebaseUserRepository.addUser(map, uid)
-
-    assert(result is Resource.Failure)
-    assert((result as Resource.Failure).throwable.message == message)
-  }
+  //  @Test
+  //  fun `test addUser(Map, String)`() = runTest {
```
Viccha18, line 219: Why did you commented these tests ? Would be nice to specify with a comment at the top

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanFriendQrViewModel.kt
```diff
@@ -0,0 +1,136 @@
...
+  fun resetNavigationEvent() {
+    _action.value = Action.None
+  }
+
+  fun changeTabState(tab: TAB) {
+    _tabState.value = tab
+  }
+
+  fun changeAction(action: Action) {
+    println("Changing action to $action")
```
Viccha18: `Log.d` would be more appropriate to display that information into the logCat when the app is running

Viccha18 (commented):
Woaw ! That's some impressive work with non trivial functionalities like camera preview implementation and QRcode decoding ! Good job ! 

I just have a few remarks and questions before I approve

app/src/main/java/com/github/se/eventradar/ui/qrCode/QrCodeFriendUi.kt
wisdomseeker2001: DONE, just not going to screen desired as not implemented yet

app/src/androidTest/java/com/github/se/eventradar/qrCode/qrCodeScanFriendUiTest.kt
```diff
@@ -0,0 +1,219 @@
...
+  private lateinit var qrCodeAnalyser: QrCodeAnalyser
+  private val myUID = "user1"
+
+  @Before
+  fun testSetup() {
+    MockKAnnotations.init(this)
+    every { mockNavActions.navigateTo(any()) } just Runs
+    userRepository = MockUserRepository()
+    (userRepository as MockUserRepository).updateCurrentUserId(myUID)
+    qrCodeAnalyser = mockk<QrCodeAnalyser>(relaxed = true)
```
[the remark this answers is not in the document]
wisdomseeker2001: I mean different uses for different things, here this is just doing a mocked relax true which means all functions we have not precise functionality of will return 0, null or false.

app/src/main/java/com/github/se/eventradar/model/di/StringDatabaseModule.kt
```diff
@@ -0,0 +1,3 @@
+package com.github.se.eventradar.model.di
+
+class StringDatabaseModule {}
```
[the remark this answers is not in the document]
wisdomseeker2001: thanks for spotting this, not needed so will delete

app/src/main/java/com/github/se/eventradar/model/repository/user/FirebaseUserRepository.kt
```diff
@@ -136,28 +136,37 @@ class FirebaseUserRepository(db: FirebaseFirestore = Firebase.firestore) : IUser
...
+            "qrCodeUrl" to user.qrCodeUrl,
+            "username" to user.username,
+            "accountStatus" to user.accountStatus,
+            "eventsAttendeeList" to user.eventsAttendeeSet,
+            "eventsHostList" to user.eventsHostSet,
+            "friendsList" to user.friendsSet,
+        )
+
+    return Pair(publicMap, privateMap)
+  }
```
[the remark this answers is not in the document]
wisdomseeker2001: sorry I didn't specify but this should not be in this PR, I had to copy and paste it from Nathan In order to test my code. once, pull from main once Nathan has pushed, It will be there

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/QrCodeAnalyser.kt
wisdomseeker2001: you are right about inject, took it off

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanFriendQrViewModel.kt
```diff
@@ -0,0 +1,135 @@
...
+  private var myUID = ""
+
+  private val _decodedResult = MutableStateFlow<String?>(null)
+  val decodedResult: StateFlow<String?> = _decodedResult.asStateFlow()
+
+  private val _action = MutableStateFlow(Action.None)
+  val action: StateFlow<Action> = _action.asStateFlow()
+
+  private val _tabState = MutableStateFlow(TAB.MyQR)
+  val tabState: StateFlow<TAB> = _tabState.asStateFlow()
```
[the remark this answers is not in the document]
Viccha18: inside

app/src/test/java/com/github/se/eventradar/FirebaseUserRepositoryUnitTest.kt
wisdomseeker2001: due to some bug ... will be hopefully resolved once I pull form Nathans merge

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanFriendQrViewModel.kt
```diff
@@ -0,0 +1,135 @@
...
+
+  enum class Action {
```
jgeaso1266, line 37: this should be in a uiState variable (like all the other view models).

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanFriendQrViewModel.kt
```diff
@@ -0,0 +1,135 @@
...
+    None,
+    NavigateToNextScreen,
+    FirebaseFetchError,
+    FirebaseUpdateError,
+    AnalyserError,
+    CantGetMyUID
+  }
+
+  enum class TAB {
```
jgeaso1266, line 46: nit: `Tab` should just have first letter capitalized.

jgeaso1266 (commented):
A few comments and questions (some which weren't answered from my previous review), but one important thing. Your view model for QR code needs to have a uiState, not two separate flows. This one uiState will be used in the UI to handle all logic between UI and view model. See the other view models for an example.

app/src/androidTest/java/com/github/se/eventradar/qrCode/qrCodeScanFriendUiTest.kt
```diff
@@ -0,0 +1,220 @@
...
+    onComposeScreen<QrCodeScanFriendUiScreen>(composeTestRule) {
+      viewModel.changeAction(ScanFriendQrViewModel.Action.NavigateToNextScreen)
+      composeTestRule.waitForIdle()
+      verify { mockNavActions.navigateTo(any()) }
+    }
+  }
+}
+
+//  @Test
+//  fun displaysAllComponentsCorrectly_CameraPermittedOnce(): Unit = run {
```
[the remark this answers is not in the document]
wisdomseeker2001: Yes for now they will be I guess, because test coverage is enough

app/src/main/java/com/github/se/eventradar/ui/qrCode/QrCodeFriendUi.kt
```diff
@@ -0,0 +1,154 @@
...
+
+// TODO cleaner code for Navigation and to correct screen
+
+@Composable
+fun QrCodeScreen(
+    viewModel: ScanFriendQrViewModel = hiltViewModel(),
+    navigationActions: NavigationActions
+) {
+
+  val qrScanUiState = viewModel.uiState.collectAsStateWithLifecycle().value
```
jgeaso1266, line 46: you should use the the state returned from `collectAsStateWithLifecycle`, not the value.

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/QrCodeAnalyser.kt
jgeaso1266: Are you going to fix this?

app/src/test/java/com/github/se/eventradar/FirebaseUserRepositoryUnitTest.kt
jgeaso1266: Has Nathaniel's PR been merged? If not then please wait for that before merging. These tests are important for coverage and for testing the repo.

app/.gradle/config.properties
jgeaso1266, line 1: This file shouldn't be in version control. Can you please add it to gitignore?

app/.idea/misc.xml
jgeaso1266, line 1: > This file shouldn't be in version control. Can you please add it to gitignore?

Same with this file.

jgeaso1266 (commented):
A few comments. I also don't feel comfortable approving until Nathaniel's changes have been merged.

app/src/test/java/com/github/se/eventradar/FirebaseUserRepositoryUnitTest.kt
wisdomseeker2001: yes has been merged so all good

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanFriendQrViewModel.kt
wisdomseeker2001: done

app/.idea/misc.xml
wisdomseeker2001: should be done now

nikhen-s (changes requested):
LGTM. Just a bug where when you switch from MyQrCodeTab to the ScanQrCodeTab the Camera does not turn on, but as we talked about this in person, there is an easy fix, after that ready to approve

nikhen-s (approved):
LGTM. Little bugs can be addressed in next sprint.

Issue #176, opened 2024-04-29
Display my Qr Code View Model


wisdomseeker2001:
Create ViewModel to feed Qr Code image to the appropriate UI screen

PR #182, opened 2024-05-01
Firebase Storage Implementation with functional Profile Picture Upload

achillet3:
# Implementation of Firebase Storage
- Addition of functions to upload an to the Firebase Storage and retrieve that image 
- Upon sign up upload the chosen picture to the Firebase storage and set the profilePictureUrl document of the user to the link of that image in the database
    - If the user does not choose a picture it will automatically be set the place holder image form the Firebase storage


app/src/main/java/com/github/se/eventradar/viewmodel/AuthenticationViewModel.kt
```diff
@@ -70,6 +88,38 @@ class LoginViewModel @Inject constructor(private val userRepository: IUserReposi
...
   }
 
+  private suspend fun uploadImageAsync(
+      selectedImageUri: Uri?,
+      uid: String,
+      folderName: String
+  ): Boolean {
+    if (selectedImageUri == null) {
+      return false
+    }
```
Viccha18, line 98: Where do you set it to be the placeholder when the user don't select an image ?

app/src/main/res/values/strings.xml
```diff
@@ -24,6 +24,8 @@
     <string name="homeScreen_events">Home</string>
     <string name="user_profile">Profile</string>
     <string name="my_hosted_events">My hosting</string>
+    <string name="profile_pics_folder">Profile_Pictures</string>
+    <string name="qr_codes_folder">QR_Codes</string>
```
Viccha18, line 28: Not sure where you use these Resource String

Viccha18 (approved):
LGTM, I only have a few questions

app/src/main/java/com/github/se/eventradar/model/repository/user/FirebaseUserRepository.kt
```diff
@@ -136,6 +138,30 @@ class FirebaseUserRepository(db: FirebaseFirestore = Firebase.firestore) : IUser
...
     }
   }
+
+  override suspend fun uploadImage(
+      selectedImageUri: Uri,
+      uid: String,
+      folderName: String
+  ): Resource<Unit> {
+    val storageRef = Firebase.storage.reference.child("$folderName/$uid")
+    return try {
```
wisdomseeker2001, line 148: will ur try block ever fail with this struct?

wisdomseeker2001 (commented):
Looks good to me on a global basis

jgeaso1266:
resolver by #176

Issue #107, opened 2024-04-15
View Upcoming Events List


nikhen-s:
## Objective
- List all events you are attending in descending order You should see the Event Date, Name and Location  (Future and Passed) Name, location & Date of each event should be displayed 

## Acceptance Criteria

- [ ] This Screen accessible from the default home screen of browsing events (similar to how Jalen presented it on Figma)
- [ ] List all events you are attending in descending order You should see the Event Date, Name and Location  (Future and Passed) Name, location & Date of each event should be displayed 
- [ ] Upon Clicking on Event, screen 2.12 (Event Overview) pops up

Nath-Assa:
Resolved in #184

Issue #183, opened 2024-05-01
Refactor EventOverviewVM for Repo-Layer User ID Handling


Nath-Assa:
**Objective:**
Enhancing the architecture of our UserRepository and EventOverviewViewModel by moving the responsibility of handling the current user's ID retrieval to the repository layer. The primary goal is to decouple the view models from directly accessing the authentication state, thus adhering more closely to the principles of clean architecture.

Nath-Assa:
Resolved in #184

PR #184, opened 2024-05-01
Handle currentUserId in repo layer + View Upcoming Events

Nath-Assa:
### Key Changes:

**User Repository Updates:**
- Implemented getCurrentUserId() in IUserRepository, MockUserRepository, and FirebaseUserRepository.
- Added unit tests for the new getCurrentUserId() method to ensure correct behavior.

**ViewModel Refinements:**
- Transitioned EventOverviewViewModel state properties from var to val for immutability and introduced userLoggedIn property
- Introduced init() and checkUserLoginStatus() to manage user session state based on repository changes.
- Enhanced event handling with two new methods:
   - `setFilterDialogOpen(isOpen: Boolean)`: Manages the visibility of the filter dialog.
   - `onSearchQueryChange(newQuery: String)`: Updates the search query state.
- Expanded EventOverviewViewModelTest to include tests for the new methods and changes.

**UI Integration:**
- Adjusted LaunchedEffect in Home.kt to trigger getEvents() unconditionally on component composition instead of reacting to eventList changes.
- Updated Home.kt to display upcoming events dynamically based on user session status.
- Implemented UI tests to validate the frontend integration with the newly adjusted view models.


app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
```diff
@@ -82,17 +112,41 @@ constructor(
...
+     */
 }
 
 data class EventsOverviewUiState(
     val eventList: EventList = EventList(emptyList(), emptyList(), null),
-    var searchQuery: String = "",
-    var isFilterDialogOpen: Boolean = false,
+    val searchQuery: String = "",
+    val isFilterDialogOpen: Boolean = false,
+    // TODO: Christine, please update `radiusInputFilter` and `freeEventsFilter` to `val`.
```
Nath-Assa, line 142: @christinehahaa Take a look at this please.

app/src/main/java/com/github/se/eventradar/model/repository/user/MockUserRepository.kt
```diff
@@ -5,6 +5,7 @@ import com.github.se.eventradar.model.User
 
 class MockUserRepository : IUserRepository {
   private val mockUsers = mutableListOf<User>()
+  var currentUserId: String? = null // Simulate current user ID
```
wisdomseeker2001, line 8: This could be made private

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
```diff
@@ -132,12 +135,12 @@ fun HomeScreen(
 
     SearchBarAndFilter(
         searchQuery = uiState.searchQuery,
-        onSearchQueryChange = { uiState.searchQuery = it },
+        onSearchQueryChange = { newQuery -> viewModel.onSearchQueryChange(newQuery) },
```
wisdomseeker2001, line 138: good respect of MVVM model. Nice work

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
```diff
@@ -40,38 +46,63 @@ constructor(
...
-                                events.data, events.data, _uiState.value.eventList.selectedEvent))
-              }
-              is Resource.Failure -> {
-                Log.d("EventsOverviewViewModel", "Error getting events for $uid")
-                _uiState.value =
-                    _uiState.value.copy(eventList = EventList(emptyList(), emptyList(), null))
-              }
-            }
-          } else {
+      userRepository.getCurrentUserId().let { userIdResource ->
```
wisdomseeker2001, line 51: have you thought of perhaps putting this in a some sort of loop to treaty fetching the events? perhaps firebase bugs at the instance u attempted, I personally think it is worth reattempting.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
```diff
@@ -40,38 +46,63 @@ constructor(
...
+        val user = userResponse.data!!
+        val attendeeList = user.eventsAttendeeSet.toList()
+        if (attendeeList.isNotEmpty()) {
+          when (val events = eventRepository.getEventsByIds(attendeeList)) {
+            is Resource.Success -> {
+              _uiState.value =
+                  _uiState.value.copy(
+                      eventList =
+                          EventList(
+                              events.data, events.data, _uiState.value.eventList.selectedEvent))
```
wisdomseeker2001, line 81: did not quite get why events.data is there 2x

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
```diff
@@ -40,38 +46,63 @@ constructor(
...
         }
       }
+      is Resource.Failure -> {
+        Log.d("EventsOverviewViewModel", "Error fetching user document")
+        _uiState.value = _uiState.value.copy(eventList = EventList(emptyList(), emptyList(), null))
+      }
+    }
+  }
+
+  fun checkUserLoginStatus() {
```
wisdomseeker2001, line 101: isn't a user always need to be logged In for our V1? This is defo good to forsee updates of app in long run

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
wisdomseeker2001, line 1: only thing I could suggest u add is some ENUM that enums different states the viewmodel goes through. e.g. Enum State { FbFetchErr, FbUpdateErr ...} and in your tests you can then test that a state flow observing this enum changed In the predicted sequence of states.

Assume x is a Stateflow, you can do x.take(n).toList where n is number of stats you expected it to change to in the test you called and then you can assert it was = to Listen (State1, state2 ,...) This is good practice as It will ensure that the flow of code is exactly as you expected

app/src/main/java/com/github/se/eventradar/model/repository/user/MockUserRepository.kt
wisdomseeker2001, line 1: overall really good work, this is super useful for testing now that you can orchestrate custom UserIds

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
nikhen-s, line 101: I think this function right now looks more like updating the user's log in status. It would be nice upon Resource.Failure to Log a statement saying that no user is logged in for easy testing too.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
```diff
@@ -82,17 +113,41 @@ constructor(
   fun onViewListStatusChanged(state: MutableStateFlow<EventsOverviewUiState> = _uiState) {
     state.value = state.value.copy(viewList = !state.value.viewList)
   }
+
+  fun setFilterDialogOpen(isOpen: Boolean) {
```
nikhen-s, line 117: Just fyi, you don't need to change this because I've already told christine, but credits to [Jalen's comment](https://github.com/swent-sp-2024-event-radar/event-radar-app/pull/130#discussion_r1585952971) - this doesn't need to take in the boolean. Whenever it's called, we know the new value will be the opposite of what it is now.

app/src/test/java/com/github/se/eventradar/EventOverviewViewModelTest.kt
```diff
@@ -131,7 +131,8 @@ class EventsOverviewViewModelTest {
 
     userRepository.addUser(mockUser)
     // MockUser is on the attendeeList for events with id "1" and "2"
-    viewModel.getUpcomingEvents("user1")
+    (userRepository as MockUserRepository).updateCurrentUserId("user1")
```
nikhen-s, line 134: Is that a reason why you casted this as MockUserRepository? I think it's been setup as `userRepository = MockUserRepository()` @Before the tests run, curious to hear your reason.

app/src/test/java/com/github/se/eventradar/MockUserRepositoryUnitTest.kt
nikhen-s, line 1: Tests LGTM :)

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
```diff
@@ -60,6 +58,11 @@ fun HomeScreen(
   // Ui States handled by viewModel
   val uiState by viewModel.uiState.collectAsState()
   LaunchedEffect(key1 = uiState.eventList) { viewModel.getEvents() }
+  LaunchedEffect(key1 = uiState.tab) {
+    if (uiState.tab == Tab.UPCOMING) {
+      viewModel.getUpcomingEvents()
+    }
+  }
```
nikhen-s, line 65: concern: If getUpcomingEvents() updates the uiState.eventList, and getEvents() is triggered whenever uiState.eventList changes, there's a possibility of an unintended recursion or multiple calls to getEvents().

nikhen-s (changes requested):
Some questions and nits, but nice Job on the getUserId function & tests. Upon trying to run the app on my device, clicking on the `upcomingEvents` tab still does not change the tab (it stays in Browse tab), try to look into that and add some UI tests in androidTest folder for the UpcomingEvents part (and switching tabs). Don't forget to check [sonar cloud](https://sonarcloud.io/project/issues?resolved=false&pullRequest=184&id=swent-sp-2024-event-radar_event-radar-app&open=AY80hbqL0-tUbcqyDyHT) to address the maintainability issues too.

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
```diff
@@ -132,12 +135,12 @@ fun HomeScreen(
 
     SearchBarAndFilter(
         searchQuery = uiState.searchQuery,
-        onSearchQueryChange = { uiState.searchQuery = it },
```
christinehahaa, line 135: Search and filter was my job, so I have made these changes already in my incoming PR. But since you've created tests for them already, let's keep them here. 

I will update your functions/tests related to search and filter with my work where necessary, since I have already tested on my end to ensure it works.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
christinehahaa: EventList contains allEvents and FilteredEvents. First one is allEvents, second is FilteredEvents. When data is first fetched, FilteredEvents =  allEvents.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
```diff
@@ -82,17 +113,41 @@ constructor(
...
+    _uiState.update { currentState -> currentState.copy(isFilterDialogOpen = isOpen) }
+  }
+
+  fun onSearchQueryChange(newQuery: String) {
+    _uiState.update { currentState -> currentState.copy(searchQuery = newQuery) }
```
christinehahaa, line 122: I will edit these functions with my work if needed, once I push my PR.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
```diff
@@ -82,17 +113,41 @@ constructor(
...
+          currentState.copy(radiusInputFilter = radius)
+      }
+  }
+  fun setFreeEventsFilter(isFree: Boolean) {
+      _uiState.update { currentState ->
+          currentState.copy(freeEventsFilter = isFree)
+      }
+  }
+
+   */
```
christinehahaa, line 136: Remove this TO DO, I've implemented already. Thank you for adding them tho :)

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
christinehahaa: Looks good, let's keep this in your PR since you've done tests for it already. That way the work is associated to you. 

I will change all my other UiState values I added to be val and not var.

app/src/test/java/com/github/se/eventradar/EventOverviewViewModelTest.kt
```diff
@@ -199,4 +218,48 @@ class EventsOverviewViewModelTest {
...
+  @Test
+  fun testOnSearchQueryChange() = runTest {
+    val newQuery = "sample search"
+    viewModel.onSearchQueryChange(newQuery)
+    assert(newQuery == viewModel.uiState.value.searchQuery)
+
+    val anotherQuery = "another search"
+    viewModel.onSearchQueryChange(anotherQuery)
+    assert(anotherQuery == viewModel.uiState.value.searchQuery)
+  }
```
christinehahaa, line 240: Thanks for adding these :)

christinehahaa (commented):
Some comments regarding code overlap with my PR, but otherwise I tested on my machine and looks good overall. Line coverage just needs to be boosted to above 80% before merging into main.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
Nath-Assa: I believe this works fine. Don't think changing this would be time-efficient.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
Nath-Assa: Yes, a User should always be logged in. This will be useful long term.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
Nath-Assa: Would be nice indeed. Something to look into during next sprint definitely.

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
```diff
@@ -167,9 +170,34 @@ fun HomeScreen(
...
+        EventList(
+            events = uiState.eventList.allEvents,
+            modifier =
+                Modifier.testTag("eventList").fillMaxWidth().constrainAs(eventList) {
+                  top.linkTo(searchAndFilter.bottom, margin = 8.dp)
+                  start.linkTo(parent.start)
+                  end.linkTo(parent.end)
+                })
+      }
       viewModel.onTabChanged(Tab.BROWSE)
```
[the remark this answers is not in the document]
Nath-Assa: yes! this is why the upcoming tab wasn't working. Will remove it.

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
```diff
@@ -167,9 +170,34 @@ fun HomeScreen(
...
+      if (uiState.eventList.allEvents.isEmpty() && uiState.userLoggedIn) {
+        Text(
+            "You have no upcoming events",
+            modifier =
+                Modifier.constrainAs(eventList) {
+                  top.linkTo(searchAndFilter.bottom, margin = 8.dp)
+                  start.linkTo(parent.start)
+                  end.linkTo(parent.end)
+                })
+      } else if (!uiState.userLoggedIn) {
```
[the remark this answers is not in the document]
Nath-Assa: In the current implementation, the scenario where a user attempts to access 'Upcoming Events' without being logged in should indeed be impossible due to the app's navigation guards. However, including this check and displaying a relevant message ('Please log in') serves as an additional layer of error handling.

Though ideally, users should never see this message under normal circumstances, it acts as a fail-safe to prompt users towards corrective action (like logging in) if they somehow navigate to this state.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
Nath-Assa: I believe Christine said that I would be keeping this method in this PR. Will edit.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
Nath-Assa: This function invokes getCurrentId(), which already handles logging a message if a user is not signed in. Including an additional log message here would be redundant.

app/src/test/java/com/github/se/eventradar/EventOverviewViewModelTest.kt
Nath-Assa: Yes, the method updateCurrentUserID() is exclusive to the MockUserRepository and is not defined in the IUserRepository interface. As a result, casting is necessary to access this method, since it would not be available through the interface otherwise.

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
Nath-Assa: Yes. Will update this.

Nath-Assa:
Still need to: 
- Add extra UI tests
- Add tests for getCurrentUserId() in FirebaseUserRepository
- Complete the upgrade of HostedEventsViewModel to align with the current user ID handling.

christinehahaa (approved):
Upcoming events tab is working now -  nice job!

For next time - it appears there is a lot of changes being made in this PR, some related, some unrelated to each other. I would split into multiple PRs so they can be pushed into main faster next time :)

Nath-Assa:
**Note:** The implementations of EventList and EventMap in Home.kt may seem to be near-duplicate code. However, each usage is contextually unique with specific test tags tailored to different UI scenarios. While extracting these into private functions could reduce code duplication superficially, it would also add complexity. After careful consideration, it was decided to keep these implementations separate to maintain code clarity and simplicity.

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
```diff
@@ -169,21 +172,58 @@ fun HomeScreen(
             })
       }
     } else {
-      // "Upcoming" tab content
-      // TODO: Implement upcoming events
-      Toast.makeText(context, "Upcoming events not yet available", Toast.LENGTH_SHORT).show()
-      viewModel.onTabChanged(Tab.BROWSE)
+      if (uiState.eventList.allEvents.isEmpty() && uiState.userLoggedIn) {
```
nikhen-s, line 175: Do you think we should still show Map View even if there are no Upcoming Events available? The map will be empty, because i think this might be what users would expect upon clicking the MapView, but we can ask others how they think about this too.

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
```diff
@@ -59,8 +56,8 @@ fun HomeScreen(
 ) {
   // Ui States handled by viewModel
   val uiState by viewModel.uiState.collectAsState()
-  LaunchedEffect(key1 = uiState.eventList) { viewModel.getEvents() }
-  val context = LocalContext.current
+
+  LaunchedEffect(Unit) { viewModel.getEvents() }
```
nikhen-s, line 60: nit: maybe it'a better to initialize this in the viewModel using `init{}` for separation of concerns?

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
```diff
@@ -40,38 +46,63 @@ constructor(
     }
   }
 
-  fun getUpcomingEvents(uid: String) {
+  fun getUpcomingEvents() {
```
nikhen-s, line 49: Is there a particular reason you removed uid here?

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
```diff
@@ -132,12 +135,12 @@ fun HomeScreen(
...
+        onSearchQueryChange = { newQuery -> viewModel.onSearchQueryChange(newQuery) },
         onSearchButtonClick = {
           // Handle search button click
         },
         showFilterPopUp = uiState.isFilterDialogOpen,
-        setShowFilterPopUp = { uiState.isFilterDialogOpen = it },
+        setShowFilterPopUp = { viewModel.changeFilterDialogOpen() },
         modifier =
             Modifier.padding(horizontal = 16.dp, vertical = 8.dp)
                 .testTag("searchBarAndFilter")
```
nikhen-s, line 146: small nit: instead of if (getTabIndexFromTabEnum(uiState.tab) == 0), maybe uiState.tab == Tab.Browsing would be more readable

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
```diff
@@ -40,38 +46,63 @@ constructor(
...
                 _uiState.value.copy(eventList = EventList(emptyList(), emptyList(), null))
           }
         }
-        is Resource.Failure -> {
-          Log.d("EventsOverviewViewModel", "Error fetching user document")
+      }
+    }
+  }
+
+  private suspend fun getUserEvents(uid: String) {
```
nikhen-s, line 69: small nit: i know that we separated hosting viewmodel and eventsOverviewViewModel, but maybe getUserUpcomingEvents is more clear so it does not get confused with getUserHostedEvents.

nikhen-s (changes requested):
Overall looks great, but just need to address 2 things and after it LGTM:

1. Try to fix this [maintainability issue](https://sonarcloud.io/project/issues?resolved=false&pullRequest=184&id=swent-sp-2024-event-radar_event-radar-app&open=AY80hbqL0-tUbcqyDyHT&tab=code) in SonarCloud to help increase the code quality. (It would also help to reduce the code dupliaction, suggestion: I put the if statements through ChatGPT and it helped suggest some ways to reuse the code & reduce the duplication)
2. (Smaller question) Should we display the MapView even if there are no Upcoming Events?

Other comments are just nits

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
Nath-Assa: This change is part of a broader update where user identification is handled directly within the repository layer, specifically using the new getCurrentUserId() method from the IUserRepository interface.

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
Nath-Assa: LaunchedEffect with Unit as the key ensures that getEvents() is called once when the composable enters the composition and not again unless the composable is recomposed from scratch. 
Triggering data loading in the ViewModel's init{} block can lead to premature requests, misaligned lifecycles, and unnecessary resource consumption.

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
Nath-Assa: Good point! I agree this deserves further discussion with the team to ensure it aligns with user expectations. We can consider making adjustments in the next sprint based on our findings.

nikhen-s (approved):
LGTM

Nath-Assa:
Closes #107 and #183

Nath-Assa:
Note: Last 5 commits can be ignored was trying to get the CI to work.

Nath-Assa:
**Detected small bug:**  When you are on Upcoming tab and press on the nav bar to navigate to another page, then come back to home, Upcoming tab shows all events(instead of upcoming). This is caused by:
`LaunchedEffect(Unit) { viewModel.getEvents() }`
Will fix this in another PR.

PR #187, opened 2024-05-01
Fix Signing Process

jgeaso1266:
## Key Changes
- Remake the APK signing CI to use the correct signing keys

Example of a [successful run](https://github.com/swent-sp-2024-event-radar/event-radar-app/actions/runs/8929205752)


nikhen-s (approved):
LGTM

PR #188, opened 2024-05-02
Bug Fix: Cannot type "/" in DOB sign up field

### LINKED ISSUE #189: Bug Fix: Cannot type "/" in DOB sign up field
Bug: Cannot type a "/" in the DOB field on sign up due to numeric keyboard

Easy fix is to removed numeric keyboard from DOB field since checks are already in place for correct DOB format.

christinehahaa:
Bug:
- Cannot type a "/" in the DOB field due to numeric keyboard

Main change:
- Removed numeric keyboard from DOB field

Closes #189

### DIFF
--- a/app/src/main/java/com/github/se/eventradar/ui/login/SignUpScreen.kt
+++ b/app/src/main/java/com/github/se/eventradar/ui/login/SignUpScreen.kt
@@ -237,7 +237,6 @@ fun SignUpScreen(
                           unfocusedPrefixColor = MaterialTheme.colorScheme.primary),
               shape = RoundedCornerShape(12.dp),
               isError = uiState.birthDateIsError,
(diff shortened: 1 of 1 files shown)


Viccha18 (approved):
That's one good idea to fix this quickly 👍  however, I would prefer that we change the parsing from "dd/mm/yyyy" to "dd.mm.yyyy" in the `isValidDate()` function of the AuthenticationViewModel. Because the "." is available in the `KeyboardType.Number` and it is more convenient for the user to have a restricted keyboard in that situation.
It would also make it consistent with the other dates in the app (in eventDetails for exemple) which are display in the same format : 21.06.2024

I won't request changes because M2 is just around the corner and it is a valid way to fix the problem. Let me know what you think, should we make a quick issue about this for next sprint ?

PR #132, opened 2024-04-18
Event homepage search bar and filter - UI

christinehahaa:
Main changes:
- added a row for the search bar and filter element
- created the filter pop up box, currently the pop up box is always shown

To be completed:
- Functioning search bar that updates when typing
- Filter button opens and closes the pop up when clicked
- UI integrated with viewmodel & updates filtered list based on search query
- UI integrated with viewmodel & updates filtered list based on filter query

PR #190, opened 2024-05-02
Home search bar and filter functionality + tests

### LINKED ISSUE #88: Create search bar and filter button: Backend
## Objective
Create a search bar and filter button/pop-up to filter the list on the browse/upcoming events screen on both map and list

## Acceptance Criteria
- [ ] In the backend, filter the list based on the search bar criteria (which would filter by name) or the filter button (which would filter by distance, category, etc.). (examples in VCS)
- [ ] Create tests for searching and filtering function

### LINKED ISSUE #117: Create search bar and filter button: ViewModel + Query the database interface + test
## Link to User Stories
#15, #16, In Figma user flows 3, 4

## Objective
Create ViewModel for Search Bar and Filter Button. 

## Acceptance Criteria
- [ ]

christinehahaa:
Related to PR #132 and closes #117 and closes #88 

## Goal
- To implement a functional search bar that filters events based on their name
- To implement a functional filter pop up that filters events based on their radius distance away from user's current location, whether free events only or not, and event category

## Main Changes
- Made updates from Nicolas (coach) review in PR #132 which include:
   - Moving strings to `strings_event.xml` file
   - Moving all state changes to the VM and not the UI
   - Updated the EventsOverviewUiState() from `var` to `val`
- Search bar functionality
   - Search bar query updates on type
   - Event cards displayed are filtered as search bar query is updated (via `filterEvents()` function in the VM)
- Filter button functionality
   - Filter button opens and closes the filter pop up
   - Radius query updates on type
   - Free events slider changes on click
   - Categories can be ticked or unticked
   - Clicking the `Apply` button in the filter pop up filters and displays the events according to all 3 filters (radius, free, category)
- Both the search bar and filter pop up filter the events simultaneously via "intersect" of filters implementation logic

**Note**: This PR was primarily worked on by Christine. Nathaniel only assisted with changes at the end to incorporate upcomingEventList as a field in the EventsOverviewUiState, as per PR #250


Nath-Assa (changes requested):
Great work, well written code!
I ran the app and found two small bugs. 
1. When the filter pop-up is left open and the map view is selected, the pop-up remains on screen without an option to close it, as the filter button disappears in this view.
2. After applying certain filters and closing the filter pop-up, reopening the pop-up shows all filters reset to their default states, despite the filters still being actively applied in the background. This inconsistency between the displayed and actual filter states can lead to user confusion. 

**Suggestion:** It would improve usability if the filter pop-up automatically closed when starting to type in the search bar, ensuring a smoother user experience and preventing overlay clutter.

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
```diff
@@ -152,6 +155,18 @@ fun HomeScreen(
 
     if (uiState.tab == Tab.BROWSE) {
```
nikhen-s, line 156: imp: I think we could reset the searchBar state when we switch to the Upcoming tab so that after switching tabs, the text in search bar is also reset :)

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
```diff
@@ -30,6 +36,118 @@ constructor(
...
+          if (isFreeSwitchOn) {
+            it.ticket.price == 0.0
+          } else {
+            true
+          }
+        }
+
+    // Filter based on categories selected
+    val categoriesCheckedList = _uiState.value.categoriesCheckedList
+    val filteredEventsCategory =
```
nikhen-s, line 115: imp: I think all the categories should be set to false at first, and in that case, no filter is applied so all the categories are shown (hopefully this helps with bug 1 of filter)

nikhen-s (changes requested):
Nice job on this :)

The searching works well in my end, but I noticed some minor bugs, and here is how to recreate them:

1. If I open the filter and deselect all of the checkboxes, then click apply, it removes any event from the screen. But I think the expected behavior is for it to show all of the events (because no filter is being applied). (included a comment as to where I think the bug might come from). 

2. When I type in some text in the search bar, and switch to a different tab, (Browse to Upcoming), the text in the search bar stays there, but I think it's supposed to reset when changing to Upcoming Tab. (I included a comment as to where i think the issue might come from).

Final thing (but future changes): These could be in a separate PR, but what do you think of creating a separate searchBar viewModel so that it can be reused it HostingEvents? (Otherwise we have to copy paste the code in HostingEvents, but i think it might be easier to have 1 viewModel so that we can fix any possible future bug just from within the searchBar viewModel).

christinehahaa:
> * If I open the filter and deselect all of the checkboxes, then click apply, it removes any event from the screen. But I think the expected behavior is for it to show all of the events (because no filter is being applied). (included a comment as to where I think the bug might come from).
> *

All bugs fixed!

christinehahaa:
> * If I open the filter and deselect all of the checkboxes, then click apply, it removes any event from the screen. But I think the expected behavior is for it to show all of the events (because no filter is being applied). (included a comment as to where I think the bug might come from).
> *

1. Actually that was the intention. The initial state is that all categories are selected, since all events are displayed. Then you can deselect or select as you wish. No categories selected means no events. I'm trying to think how to make this less confusing. Or should both "no categories selected" and "all categories selected" display all events?
2. Fixed

Yes, good idea to make searchBarAndFilter into a new VM. I have created an issue out of this to put into the scrum and informed @Viccha18 (SM).

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
christinehahaa: Done!

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
christinehahaa: Ok, I've set all categories to false first.

app/src/main/java/com/github/se/eventradar/ui/component/ReusableComponents.kt
```diff
@@ -144,162 +146,170 @@ fun EventCard(event: Event, onCardClick: (String) -> Unit) {
...
       Row(
           verticalAlignment = Alignment.CenterVertically,
           horizontalArrangement = Arrangement.Start,
-          modifier = Modifier.padding(vertical = 0.dp)) {
+          modifier = Modifier.padding(vertical = 0.dp).testTag("categoryRow")) {
             Checkbox(
                 checked = isChecked,
-                onCheckedChange = { isChecked = it },
-                modifier = Modifier.scale(0.6f).size(10.dp).padding(start = 10.dp))
+                onCheckedChange = {
```
nikhen-s, line 296: Would be nice to add a test covering this, but other than that, LGTM!

nikhen-s (approved):
LGTM!

app/src/test/java/com/github/se/eventradar/EventOverviewViewModelTest.kt
```diff
@@ -167,6 +168,7 @@ class EventsOverviewViewModelTest {
             listOf(
                 mockEvent.copy(eventName = "Event 1", fireBaseID = "1"),
                 mockEvent.copy(eventName = "Event 2", fireBaseID = "2")))
+    println(viewModel.uiState.value.upcomingEventList.filteredEvents.size)
```
christinehahaa, line 171: Probably can delete this print statement, as it's only used for debugging?

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
Nath-Assa, line 1: Nit: Cognitive complexity is 23. It is recommended to have a score <=15.

app/src/main/java/com/github/se/eventradar/ui/component/ReusableComponents.kt
```diff
@@ -144,162 +146,170 @@ fun EventCard(event: Event, onCardClick: (String) -> Unit) {
...
+                  onClick = { onFilterApply() }, modifier = Modifier.testTag("filterApplyButton")) {
+                    Text(stringResource(id = R.string.filter_apply))
+                  }
+            }
       }
     }
   }
 }
 
+// TO DO: Need to convert category selection to MVVM and test
```
Nath-Assa, line 284: This method is not tested. If you don't plan on testing it in the future, you can remove the comment.

app/src/main/java/com/github/se/eventradar/ui/component/ReusableComponents.kt
```diff
@@ -144,162 +146,170 @@ fun EventCard(event: Event, onCardClick: (String) -> Unit) {
...
-          modifier = Modifier.padding(vertical = 0.dp)) {
+          modifier = Modifier.padding(vertical = 0.dp).testTag("categoryRow")) {
             Checkbox(
                 checked = isChecked,
-                onCheckedChange = { isChecked = it },
-                modifier = Modifier.scale(0.6f).size(10.dp).padding(start = 10.dp))
+                onCheckedChange = {
+                  isChecked = it
+                  if (isChecked) {
+                    uiState.categoriesCheckedList.add(category)
```
Nath-Assa, line 299: Instead of modifying the categoriesCheckedList directly within the composable function, the UI should call a function in the ViewModel that handles this logic. This will improve our adherence to MVVM principles.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
```diff
@@ -108,31 +240,34 @@ constructor(
...
     val searchQuery: String = "",
+    val isSearchActive: Boolean = false,
     val isFilterDialogOpen: Boolean = false,
-    var radiusInputFilter: Double = -1.0,
-    var freeEventsFilter: Boolean = false,
-    val categorySelectionFilter: List<EventCategory> = emptyList(),
+    val isFilterActive: Boolean = false,
+    val radiusQuery: String = "",
+    val isFreeSwitchOn: Boolean = false,
+    val categoriesCheckedList: MutableSet<EventCategory> = mutableSetOf(),
```
Nath-Assa, line 270: In MVVM architecture, it’s typically better to keep the UI state as immutable as possible to prevent unwanted side effects. Consider changing this val to `Set<EventCategory>`.

Nath-Assa (commented):
Overall, great work! A few suggestions.

app/src/main/java/com/github/se/eventradar/ui/component/ReusableComponents.kt
Nath-Assa, line 1: Unused import in line 25:  `import androidx.compose.material.icons.filled.List`

Note: can't comment directly on the line as it's not a change from this PR.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
```diff
@@ -30,6 +36,130 @@ constructor(
     checkUserLoginStatus()
   }
 
+  fun onSearchQueryChanged(
```
Nath-Assa, line 39: **Question:** Why is _uiState passed as a default parameter? It doesn't seem necessary. 

Note: This question applies to all functions in the file that use _uiState as a default parameter.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
```diff
@@ -30,6 +36,130 @@ constructor(
...
+  fun onFreeSwitchChanged(state: MutableStateFlow<EventsOverviewUiState> = _uiState) {
+    state.value = state.value.copy(isFreeSwitchOn = !state.value.isFreeSwitchOn)
+  }
+
+  fun onFilterApply(state: MutableStateFlow<EventsOverviewUiState> = _uiState) {
+    state.value = state.value.copy(isFilterActive = true)
+    filterEvents()
+  }
+
+  fun filterEvents() {
```
Nath-Assa, line 75: **Suggestion:** The filterEvents() method could be improved by combining the multiple filtering operations into a single pass over the event list. Currently, the method filters the list separately for search queries, radius, free events, and categories, which is not efficient.

Additionally, the decision logic to determine the list based on the Tab is split between the start and end of the method. It would be cleaner to handle this selection only once.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
Nath-Assa, line 270: **Suggestion:** The current implementation of the category filter in the UI behaves such that no categories are selected by default, which effectively shows events from all categories. However, manually selecting all categories leads to the same result. This doesn't allow us to select no categories and could potentially confuse users about the filter's function. To improve user experience and clarity, consider initializing the application with all categories checked by default.

app/src/main/java/com/github/se/eventradar/ui/component/ReusableComponents.kt
```diff
@@ -144,162 +146,170 @@ fun EventCard(event: Event, onCardClick: (String) -> Unit) {
...
 
+// TO DO: Need to convert category selection to MVVM and test
 @Composable
-fun CategorySelection() {
-  LazyColumn {
+fun CategorySelection(uiState: EventsOverviewUiState, modifier: Modifier) {
+  LazyColumn(modifier = modifier) {
     items(EventCategory.entries) { category ->
-      var isChecked by remember { mutableStateOf(false) }
+      var isChecked by remember { mutableStateOf(uiState.categoriesCheckedList.contains(category)) }
```
Nath-Assa, line 289: Linked to my other comment on no categories selected by default.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
christinehahaa: Not sure when the user would want to see "no events", which is why both "no categories selected" and "all categories selected" produces the same result being "all events". I have decided to keep it as is, as per Nik's suggestion.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
christinehahaa: It means that when the function is called in the UI (view), it doesn't need to pass in any parameters. We know that the UI will always act on the same viewmodel, explicitly the default UiState specified in the EventsOverviewViewModel. So it is cleaner in the UI to pass _uiState as a default parameter.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
christinehahaa: I have made filterEvents more efficient by combining all filters into one. 

Regarding the tab split logic, it is not possible to do this only once without making the code even less efficient. i.e. this is the most efficient way to do this. Other means would require repeating logic.

app/src/main/java/com/github/se/eventradar/ui/component/ReusableComponents.kt
christinehahaa: I have already spent a few hours trying to test this part but with no success. Since test coverage is above 80% and testing this feature on the emulator seems to work already, I think it's more important to get this PR merged so I can start spending my time on other features (delivering value) even if this PR is not perfect, rather than focusing my time on testing every single part of the code.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
christinehahaa: Since this Set is in the ViewModel, it can be mutable. It only needs to be immutable in the UI. i.e. the view should not change the state.

app/src/main/java/com/github/se/eventradar/ui/component/ReusableComponents.kt
christinehahaa: Hi Nathaniel, I considered this issue last week too. My approach was to follow what was done for the freeSwitch and RadiusQuery and move the logic into a function in the VM. I kept isChecked the same. This did not work however. Then, I also tried to make isChecked into a Map<EventCategory, Boolean> so we could store the checked state inside EventOverviewUiState. This also did not work. I spent 4 hours last week attempting these approaches and another 2 hours again today. Both attempts have been unsuccessful. 

I have also tried to search online for solutions that use checkbox and follow "MVVM". All solutions seem to use the current working approach I have of tracked isChecked in a mutableStateOf. 

Given that devs are expected to spend 10 hours each week coding, and we have many other features to implement before M3, it may be best to prioritise other features/issues over this one. This will ensure that we continue to deliver value as a whole. 

If you have an other suggested approaches to fix this issue, I am happy for you to try them. Otherwise, we can create a new issue out of this on the SCRUM in our PB.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
christinehahaa: See prior comment.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
Nath-Assa: I meant: Why not directly use the _uiSate without passing it as a parameter, since we are already inside EventsOverviewVM?

Nath-Assa (approved):
Approved for merge. Although we could refine the code to adhere more closely to best practices, it functions as intended for now. To maintain efficiency, we'll proceed with merging this PR. We can always revisit and further optimize in a future PR if time allows.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
christinehahaa: It means you don't have to pass it as a parameter in the UI

Issue #175, opened 2024-04-29
Display QR Code on MY QR Code Screen


wisdomseeker2001:
- Need to Display the User's QR Code on the Screen. 
- To be implemented in the UI file @wisdomseeker2001 implemented from branch QrCodeScreen

PR #221, opened 2024-05-03
MyQRCode: UI + tests and ViewModel + tests 

### LINKED ISSUE #176: Display my Qr Code View Model

nikhen-s:
Dependency: waits for #181 to be merged.
Related issue #176 + #175 => Closes #176 + #175 

## Changes
- Created a MyQrCode ViewModel + Tests
- Created a MyQRCode UI + Tests, displays the userName and the QRCode Image (still needs to be tested through running the app with a newly signed up user).


app/src/test/java/com/github/se/eventradar/MyQrCodeViewModelUnitTest.kt
wisdomseeker2001, line 1: only test i would add, is to check that states are reinitiated as you want them to be when someone goes on screen., comes off and comes back on or switches tab and comes back on... (take reset into account if u want that to happen)

wisdomseeker2001 (approved):
I approve ur changes given also. Tested locally but i advise you to make the chnages i suggested for next sprint if u dont have time today

PR #227, opened 2024-05-03
Revert "QR CODE MVVM MODEL (UPDATED)  UI + VM"

nikhen-s:
Reverts swent-sp-2024-event-radar/event-radar-app#181
- The PR is really big so it would be helpful to merge these into smaller PRs.


wisdomseeker2001 (approved):
Looks promising

PR #228, opened 2024-05-03
Friend qr code scan VM

wisdomseeker2001:
Please refer to:`[ #181](https://github.com/swent-sp-2024-event-radar/event-radar-app/pull/181)`

This is an attempt to split this PR into 2.
This is the ViewModel Part. 

Issue Faced:
In the original PR 181. Line Coverage was over 80%. However This will not be the case for this PR and can not. From our understanding, the file QrCodeAnalyser.kt  can not be tested via code via written tests, only via real simulation of the camera with an android phone to analyse a QR Code and see that it works which has been done and confirmed. 

When the UI + Tests of the Qr Scanning functionality were also part of the PR, the ratio of line uncovered was much lower given I had very high coverage of the UI file. 

It was hence decided as a team that this PR should be accepted under these unique circumstances given that we were sure the original PR passed sonarCloud conditions.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -56,10 +56,13 @@ android {
   }
 
   testOptions {
+    unitTests.isReturnDefaultValues = true
     packagingOptions { jniLibs { useLegacyPackaging = true } }
   }
 }
-
+val activityComposeVersion = "1.9.0"
+val androidXCameraVersion = "1.3.3"
+val androidXEmulatorVersion = "2.3.0"
 dependencies {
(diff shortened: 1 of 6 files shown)


Viccha18 (approved):
LGTM
This is the first half of PR : https://github.com/swent-sp-2024-event-radar/event-radar-app/pull/181 On the origine PR line coverage was above 80% 
Considering the imbalance amount of code between the two sub PR, I approve this PR despite having only 65% line coverage because it contains a components (QrCodeAnalyser) very hard to mock and achieved the 80% goal when it was packed with the UI PR

PR #232, opened 2024-05-05
Scan Ticket VM

wisdomseeker2001:
Hi, here is the scanning Ticket VM + Tests

This was actually very different to VM of scanning a friend as it is a continuous flow of scanning so need to take this into account. 

I had to change the Qr Code Analyser to invoke a different function when a Qr Code is scanned to know if we are scanning a Qr code to add a friend or to scan a ticket. 

From start to finish this took me 4.5 hours.


app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/QrCodeAnalyser.kt
```diff
@@ -19,7 +19,8 @@ class QrCodeAnalyser @Inject constructor() : ImageAnalysis.Analyzer {
   private val supportedImageFormats =
       listOf(ImageFormat.YUV_420_888, ImageFormat.YUV_422_888, ImageFormat.YUV_444_888)
 
-  var onDecoded: ((String?) -> Unit)? = null //
+  var onFriendDecoded: ((String?) -> Unit)? = null
+  var onTicketDecoded: ((String?) -> Unit)? = null
```
Viccha18, line 23: Is it really necessary to have two variables here ? One for the friend case and one for ticket case ? I see that below you use an if/else to separate the cases, can't it be just a generic var `onQrDecoded` that you call at the end of the decoding ? Since you said in comment below that it has to be reset to null by the viewModel later

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanTicketQrViewModel.kt
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  private fun updatePermissions(decodedString: String) {
+    while (myEventID == null) {
+      // Wait until myEventID is not null
+    }
+    println("entered updatePermissions")
+    val uiLength = 28
+    val attendeeID = decodedString.take(uiLength)
+    Log.d("QrCodeFriendViewModel", "Ticket User ID: $attendeeID")
```
Viccha18, line 69: nit : don't forget to change the tag to `QrCodeTicketViewModel`

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanTicketQrViewModel.kt
```diff
@@ -0,0 +1,132 @@
...
+            } else {
+              changeAction(Action.FirebaseUpdateError)
+              Resource.Failure(Exception("Failed to update user and event"))
+            }
+      } while ((updateResult !is Resource.Success) && (maxNumberOfRetries-- > 0))
+    } else {
+      println("one does not contain the other")
+      changeAction(Action.DenyEntry)
+    }
+  }
```
Viccha18, line 113: nit: typically a duplicated functionality between the two viewModel. If time, would be nice to have it as a decorator (like coaches suggested) to avoid duplication of code

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanTicketQrViewModel.kt
```diff
@@ -0,0 +1,132 @@
...
+  private fun updateDecodedString(result: String) {
+    _uiState.value = _uiState.value.copy(decodedResult = result)
+  }
+
+  fun saveEventID(eventID: String) {
+    myEventID = eventID
+  }
+
+  fun changeAction(action: Action) {
+    _uiState.value = _uiState.value.copy(action = action)
```
Viccha18, line 124: nit : same
```kotlin
_uiState.update {
    it.copy(action = action)
}
```

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanTicketQrViewModel.kt
```diff
@@ -0,0 +1,132 @@
...
+            }
+      } while ((updateResult !is Resource.Success) && (maxNumberOfRetries-- > 0))
+    } else {
+      println("one does not contain the other")
+      changeAction(Action.DenyEntry)
+    }
+  }
+
+  private fun updateDecodedString(result: String) {
+    _uiState.value = _uiState.value.copy(decodedResult = result)
```
Viccha18, line 116: nit : would be better to use
```kotlin
_uiState.update {
    it.copy(decodedResult = result)
}
```

app/src/test/java/com/github/se/eventradar/ScanTicketQrViewModelTest.kt
```diff
@@ -0,0 +1,171 @@
...
+    viewModel = ScanTicketQrViewModel(userRepository, eventRepository, qrCodeAnalyser)
+  }
+
+  @Test
+  fun testDecodingSuccess() = runTest {
+    userRepository.addUser(mockUser1)
+    eventRepository.addEvent(mockEvent1)
+    val testDecodedString = "user1"
+    qrCodeAnalyser.onTicketDecoded?.invoke(testDecodedString)
+    TestCase.assertEquals(testDecodedString, viewModel.uiState.value.decodedResult)
```
Viccha18, line 115: nit: to stay consistent with the other test, I would just use `assertEquals` no needs for the `TestCase` in front of it

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanTicketQrViewModel.kt
```diff
@@ -0,0 +1,132 @@
...
+        changeAction(Action.AnalyserError)
+        //        println("checkpoint 2")
+      }
+    }
+  }
+
+  private fun updatePermissions(decodedString: String) {
+    while (myEventID == null) {
+      // Wait until myEventID is not null
+    }
```
Viccha18, line 65: Is this needed because of what we've discussed about runtime argument not being able to be passed alongside injected argument in the viewModel constructor ?

Viccha18 (changes requested):
Good job ! Just some small corrections and questions.

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanTicketQrViewModel.kt
wisdomseeker2001: nice, will do this now
thanks

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanTicketQrViewModel.kt
wisdomseeker2001: disagree, they do different things.

Viccha18: I meant the fact to retry, I know the two functions need to fetch different things

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/QrCodeAnalyser.kt
wisdomseeker2001: Yes, I was just being over precautious ... changed it all now I am certain :)

PR #233, opened 2024-05-05
resetting states to avoid bugs 

wisdomseeker2001:
also takes into account hat the QrCode Analyser is shared by both the Ticket & Friend Scanner

### DIFF
--- a/app/src/main/java/com/github/se/eventradar/ui/qrCode/QrCodeFriendUi.kt
+++ b/app/src/main/java/com/github/se/eventradar/ui/qrCode/QrCodeFriendUi.kt
@@ -52,8 +52,7 @@ fun QrCodeScreen(
         navigationActions.navigateTo(
             TOP_LEVEL_DESTINATIONS[
                 1]) // TODO change to private message screen with friend // Adjust according to your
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/se/eventradar/ui/qrCode/QrCodeFriendUi.kt
```diff
@@ -142,7 +141,10 @@ fun QrCodeScreen(
       }
     }
     BottomNavigationMenu(
-        onTabSelected = { tab -> navigationActions.navigateTo(tab) },
+        onTabSelected = { tab ->
+          navigationActions.navigateTo(tab)
+          viewModel.resetConditions()
```
jgeaso1266, line 146: Should this run for every tab on the bottom navigation screen or only when going off of the QR Code screen? If so, then lines 145 and 146 should be switched.

jgeaso1266 (commented):
One comment. Also needs a test as you put in the TODO.

app/src/main/java/com/github/se/eventradar/ui/qrCode/QrCodeFriendUi.kt
wisdomseeker2001: thanks, will correct this

PR #237, opened 2024-05-07
Change app manifest name to `Event Radar`

jgeaso1266:
## Key Changes
- update the app manifest to have `Event Radar` as app name instead of `My Application`


Viccha18 (approved):
LGTM

PR #238, opened 2024-05-07
Force workflow to run on all release branches

jgeaso1266:
All release branches have the convention of "M*", where * corresponds to the milestone number.


Viccha18 (approved):
LGTM thank you for the fix !

PR #242, opened 2024-05-07
Update google-services.json

jgeaso1266:
## Key Changes
- redownload `google-services.json` file from Firebase console to allow authentication to work for different debug keys


nikhen-s (approved):
LGTM

PR #248, opened 2024-05-09
View model factory for EventDetailsViewModel

Viccha18:
# Objective

This aims to solve the problem of passing runtime arguments alongside injected arguments in the viewModel constructor. 

# Note

I had to update `hilt-navigation-compose to version 1.2.0`
That change can be applied to other viewModel with these requirement as well

edit : that link really helped me implementing it https://proandroiddev.com/hilt-viewmodels-assisted-injection-aca2d6ee581d


Viccha18:
I have no idea how to test this... already spent 4 hours trying to test

wisdomseeker2001:
So, 

After having looked deeply into your code, I understand what is going on, how I still have some doubts I need to clear. However the result of this implementation will be super useful and Clean for many viewModels.
The code did not have any duplication as such or things I would change. Only problem is the coverage, which perhaps I can help with once I understand better scenarios we can test.

app/src/androidTest/java/com/github/se/eventradar/event/EventDetailsViewModelFactoryTest.kt
david-schroeter, line 1: I am not exactly sure to understand what these tests are actually testing? Calling the method `factory.create(...)` and then verifying that it has indeed been called doesn't seems like a very useful test.

I saw that you already wrote a unit test to test the functionality of your `ViewModel` was there a particular reason for wanting to test it a second time here ?

If you want to explicitly call the `create` method from your companion object, I would suggest you to slightly modify your Unit test to create your `ViewModel` with it.

I provided a possible implementation in the `EventDetailsViewModelUnitTest.kt` file to achieve this.

Let me know if it helped.

app/src/test/java/com/github/se/eventradar/EventDetailsViewmodelUnitTest.kt
```diff
@@ -64,8 +64,7 @@ class EventDetailsViewmodelUnitTest {
   @Before
```
david-schroeter, line 64:
```suggestion
    private val factory = object : EventDetailsViewModel.Factory {
      override fun create(eventId: String): EventDetailsViewModel {
        return EventDetailsViewModel(eventRepository, eventId)
      }
    }

    @Before
```

app/src/test/java/com/github/se/eventradar/EventDetailsViewmodelUnitTest.kt
```diff
@@ -64,8 +64,7 @@ class EventDetailsViewmodelUnitTest {
...
   fun setUp() {
     eventRepository = MockEventRepository()
-    viewModel = EventDetailsViewModel(eventRepository)
-    viewModel.saveEventId(mockEvent.fireBaseID)
+    viewModel = EventDetailsViewModel(eventRepository, mockEvent.fireBaseID)
```
david-schroeter, line 67:
```suggestion
    viewModel = factory.create(mockEvent.fireBaseID)
```

david-schroeter (commented):
Great job on improving the `EventDetailsViewModel` class to be able to inject runtime dependencies using Hilt. It is a relatively new way of doing it like this and therefore there isn't a lot of well documented examples of how to achieve this, but the way you found seems like the best one to me.

I also spent quite some times trying to read different blogs to learn how to do it, and your version is very clean, nice job ! 👍🏻 

---

For the other to understand how it works, I put here a couples of links that have been useful to me to understand how it works, feel free to read them if you want to learn more about how it works.

- [Hilt dependency injection basics (Android doc)](https://developer.android.com/training/dependency-injection/hilt-android)
- [Passing arguments to Hilt ViewModel](https://medium.com/@cgaisl/how-to-pass-arguments-to-a-hiltviewmodel-from-compose-97c74a75f772) The most interesting chapter from this article is `The Solution — How to Pass Arguments to a HiltViewModel from Compose`.

christinehahaa (approved):
I have reviewed, as I will need this for Private Chat UI as well. After all the changes from the comments provided by David, LGTM!

Issue #220, opened 2024-05-03
Refine Event Handling on Home Screen


Nath-Assa:
**Objective:**
- Bug Fix: Correct the issue in the Home screen where navigating away from and back to the Upcoming tab incorrectly displays all events 
   instead of just upcoming ones. This is due to the LaunchedEffect(Unit) which reinitializes getEvents() each time the Home view is 
   revisited.

- When you have no upcomings events, pressing on map icon should show the map.

- Enhance the EventsOverviewUiState by introducing a new upcomingEvents list to differentiate between all events and specifically upcoming ones. This update will ensure clearer state management and optimizes UI updates specific to upcoming events, improving the responsiveness and accuracy of event displays.

Nath-Assa:
Resolved in #250

PR #250, opened 2024-05-09
Refine Event Handling on Home Screen

Nath-Assa:
### Key Changes:

- Bug Fix: Previously, navigating away from and then returning to the Upcoming tab incorrectly displayed all events instead of just the upcoming ones. This issue has been resolved.

- When you have no upcomings events, pressing on map icon now shows the map.

- Enhanced the EventsOverviewUiState by introducing a new upcoming events list to differentiate between all events and specifically upcoming ones. This update ensures clearer state management and optimizes UI updates specific to upcoming events, improving the responsiveness and accuracy of event displays.

### DIFF
--- a/app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
+++ b/app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
@@ -57,10 +57,12 @@ fun HomeScreen(
   // Ui States handled by viewModel
   val uiState by viewModel.uiState.collectAsState()
 
-  LaunchedEffect(Unit) { when{
+  LaunchedEffect(Unit) {
+    when {
       (uiState.tab == Tab.BROWSE) -> viewModel.getEvents()
       else -> viewModel.getUpcomingEvents()
-  }  }
+    }
+  }
 
   ConstraintLayout(modifier = Modifier.fillMaxSize().testTag("homeScreen")) {
     val (logo, tabs, searchAndFilter, filterPopUp, eventList, eventMap, bottomNav, viewToggle) =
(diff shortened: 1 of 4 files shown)


christinehahaa (approved):
Great - these changes will be useful for reducing code complexity and the search and filter functionality as well. I have also tested on an android. LGTM!

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
```diff
@@ -183,7 +188,7 @@ fun HomeScreen(
                       start.linkTo(parent.start)
                       end.linkTo(parent.end)
                     })
-        (uiState.eventList.allEvents.isEmpty()) ->
+        (uiState.viewList && uiState.upcomingEventList.allEvents.isEmpty()) ->
```
jgeaso1266, line 191: this is not a very good use of the when block. This logic should be split. One if/when statement for dealing with whether the list or the map is shown, and another for what will be shown based on the uiState. Having them together like this is not very readable.

jgeaso1266 (commented):
One comment, otherwise looks good.

app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
Nath-Assa: done

Nath-Assa: After checking Sonar Cloud, this change actually makes our cognitive complexity higher. Reverting the changes.

Nath-Assa:
Closes #220

Issue #252, opened 2024-05-09
Implement Real-time Updates in Home.kt part2/2


Nath-Assa:
**Objective:** Implement a real-time update mechanism using a Firebase listener to refresh the event lists automatically when new events are added or existing events are updated. This will ensure that the event lists are always current and reflective of the latest data without needing a manual refresh.

**Changes:**
- Implemented the observeAllEvents(Part 1) and observeUpcomingEvents methods in both MockEventRepository and
  FirebaseEventRepository to enable real-time data streaming from Firebase. This ensures both all("browse") and upcoming
  events are continuously updated.
- Adapted EventsOverviewViewModel to initialize and handle updates through its observeAllEvents(Part 1) and observeUpcomingEvents 
  methods, linking repository changes directly to the UI.
- Wrote unit tests for every function.
- Thoroughly tested all functionalities by running the app and manually modifying entries in the Firestore database.

Nath-Assa:
Resolved in #258

PR #258, opened 2024-05-09
Implement Real-time Updates in Home.kt

Nath-Assa:
### Key Changes:
- Implemented the observeAllEvents and observeUpcomingEvents methods in both MockEventRepository and 
  FirebaseEventRepository to enable real-time data streaming from Firebase. This ensures both all("browse") and upcoming 
  events are continuously updated.
- Adapted EventsOverviewViewModel to initialize and handle updates through its observeAllEvents and observeUpcomingEvents methods, linking repository changes directly to the UI.


Nath-Assa:
@david-schroeter @RaulinN 
I’ve been reviewing our current setup in Home.kt, particularly how changes in the event lists are managed. We currently use collectAsState() for state observation and UI recomposition. Before making any changes, I’d like to get your input on whether this approach is optimal or if adding a LaunchedEffect specifically for monitoring changes in eventList and upcomingEventList would enhance our responsiveness or efficiency.

Nath-Assa:
Note: CI currently fails but it's a known bug. I need to rerun multiple times to get it to pass. I will take care of this after the tests are implemented.

Nath-Assa:
**Testing Update** 

I have conducted extensive testing on the emulator to validate the functionality of our event observation mechanisms in the app.

Observations on observeEvents:
The listener for observeEvents is functioning correctly and reacts in real-time to changes made in the database. I successfully added and deleted events, and the changes were immediately reflected in the app without any manual intervention.
image
image
image

Issues with observeUpcomingEvents:
However, I encountered issues with the observeUpcomingEvents. As a reminder, `getUpcomingEvents()` first interacts with the user database, checking the current user's attendeeList for event IDs and retrieves the corresponding events.

During UI testing, I manually added an event to my attendee list (User ID: QKPpsHHs0NPkb5RzWgDMsL7cHQt2). Unfortunately, the changes were not automatically reflected in the UI. I had to manually switch tabs (from "Upcoming" to "Browse" and back) to see the updates. This indicates that while the getUpcomingEvents function performs as expected when triggered, the observeUpcomingEvents does not update the UI in real time as intended.

Next Steps:
Adjust the implementation of observeUpcomingEvents. Currently, it listens for changes in the attendee lists of all events. To align with the functionality of the getUpcomingEvents function, it should be modified to specifically track changes in the attendee list of the current user.

Nath-Assa:
Observations on observeEvents:
The observeEvents listener is successfully monitoring real-time changes in the database. This functionality was demonstrated when I added my user ID ("QKPpsHHs0NPkb5RzWgDMsL7cHQt2") to an event's attendees_list, which led to immediate and automatic updates in the UI. Furthermore, when I removed my user ID from an event's attendees_list, the event promptly disappeared from the UI, confirming the listener's responsiveness to both additions and deletions.

Important Note: The listener is configured to specifically monitor changes within the attendees_list of events. Therefore, modifications made directly to a user's attendees_list—such as adding or removing an event ID—do not automatically trigger UI updates.

This emphasizes the necessity for robust synchronization between a user's attendeeList and the event's attendees_list. Ensuring that any update to an event’s attendees_list (either adding or removing a user ID) triggers corresponding updates on the user’s profile is crucial for consistent UI representation and maintaining data integrity across the application. This synchronization is vital for the seamless operation of our application, ensuring that users see their event engagements accurately reflected in real-time.

Nath-Assa:
I have been struggling to write effective unit tests for the observeEvents function in the FirebaseEventRepository class. Despite my efforts, I am encountering issues with capturing and triggering the Firestore event listener correctly. 

Here is a snippet of the test code I am currently working with:
`@Test
fun `observeEvents emits events when data changes`() = runTest {
    val mockSnapshot = mockk<QuerySnapshot>()
    val mockDocument = mockk<DocumentSnapshot>()
    every { mockDocument.id } returns "event1"
    every { mockDocument.data } returns mapOf("name" to "Test Event")
    every { mockSnapshot.documents } returns listOf(mockDocument)

    val listenerSlot = slot<EventListener<QuerySnapshot>>()
    every { mockEventRef.addSnapshotListener(capture(listenerSlot)) } returns mockk()

    val flow = repository.observeEvents()
    val job = launch {
        val result = flow.first()
        when (result) {
            is Resource.Success -> {
                assertEquals(1, result.data.size)
            }
            else -> fail("Expected Resource.Success but got $result")
        }
    }

    listenerSlot.captured.onEvent(mockSnapshot, null)
    job.cancel()
}`

Specifically, I keep running into the following error:
`java.lang.IllegalStateException: Value not yet captured`

app/src/test/java/com/github/se/eventradar/EventOverviewViewModelTest.kt
```diff
@@ -440,4 +442,103 @@ class EventsOverviewViewModelTest {
...
+  fun testObserveEvents() = runTest {
+    val testEvent1 = mockEvent.copy(fireBaseID = "1", eventName = "Test Event 1")
+    val testEvent2 = mockEvent.copy(fireBaseID = "2", eventName = "Test Event 2")
+    eventRepository.addEvent(testEvent1)
+
+    // Setup your ViewModel with mocked dependencies
+    // automatically triggers the 'init' block, which calls 'observeEvents()'
+    val viewModel = EventsOverviewViewModel(eventRepository, userRepository)
+
+    eventRepository.addEvent(testEvent1)
```
jgeaso1266, line 456: why do you add this event twice?

app/src/test/java/com/github/se/eventradar/MockEventRepositoryUnitTest.kt
```diff
@@ -193,4 +196,66 @@ class MockEventRepositoryUnitTest {
...
+
+    // Add a new upcoming event that the specified user will attend
+    val newEvent =
+        mockEvent.copy(
+            fireBaseID = "2",
+            eventName = "New Upcoming Event",
+            attendeeList = mutableListOf(userId))
+    // Add a new event that the specified user will not attend
+    val newEvent2 =
+        mockEvent.copy(fireBaseID = "2", eventName = "New Event", attendeeList = mutableListOf())
```
jgeaso1266, line 246: These two events have the same firebaseID. Does this affect the test? Each event usually has a unique ID.

jgeaso1266 (commented):
Coverage is still very low, I see your comments about having trouble testing some of the functions.

app/src/test/java/com/github/se/eventradar/MockEventRepositoryUnitTest.kt
Nath-Assa: doesn't affect the tests but i'll change it.

Nath-Assa:
> Coverage is still very low, I see your comments about having trouble testing some of the functions.

After testing the function in `FirebaseEventRepository.kt`, coverage will be over 80%.

app/src/main/java/com/github/se/eventradar/model/event/Event.kt
```diff
@@ -35,7 +43,7 @@ data class Event(
       ticket =
           EventTicket(
               name = map["ticket_name"] as String,
-              price = map["ticket_price"] as Double,
+              price = convertToDouble(map["ticket_price"]),
```
RaulinN, line 46: This feels a bit strange to me. Your `convertToDouble` function takes literally *anything* and tries to convert it into a `Double`. Why do you need a special function to convert to double when you do it like this `longitude = map["location_lng"] as Double` for the longitude?

RaulinN: Note: if you **really** something similar to `convertToDouble`, I'd suggest using Kotlin's **safe casts**, i.e.,

```kotlin
val p = dataFromFirebase["my_property"] as? Type1 // Notice the "?"
```

`as? Type1` attempts to cast the value to `Type1`. However, instead of throwing a `ClassCastException` if the cast fails, `as?` returns `null`

Note that the value `p` is of type `Type1?` (nullable `Type1`)

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
```diff
@@ -203,6 +205,63 @@ constructor(
...
+          }
+          is Resource.Failure -> {
+            Log.d(
+                "EventsOverviewViewModel",
+                "Error fetching user ID: ${userIdResource.throwable.message}")
+          }
+        }
+      }
+    }
+  }
```
RaulinN, line 263: There is *a lot* of indentation in this code, up to **12** which probably sets a new record! ^^ I'd suggest reducing the indentation using modularization. It will be way easier to read and understand the flow of your code. Here is an example:

```suggestion
private fun observeUpcomingEvents() {
    viewModelScope.launch {
        val userIdResource = userRepository.getCurrentUserId()
        handleUserIdResource(userIdResource)
    }
}

private suspend fun handleUserIdResource(userIdResource: Resource<String>) {
    when (userIdResource) {
        is Resource.Success -> {
            val uid = userIdResource.data
            val eventsResource = eventRepository.observeUpcomingEvents(uid).collect()
            handleEventsResource(eventsResource)
        }
        is Resource.Failure -> {
            Log.d(
                "EventsOverviewViewModel",
                "Error fetching user ID: ${userIdResource.throwable.message}"
            )
        }
    }
}

private suspend fun handleEventsResource(eventsResource: Resource<List<Event>>) {
    when (eventsResource) {
        is Resource.Success -> {
            updateUiState(eventsResource.data)
            filterEvents()
        }
        is Resource.Failure -> {
            Log.d(
                "EventsOverviewViewModel",
                "Failed to fetch upcoming events: ${eventsResource.throwable.message}"
            )
        }
    }
}

private fun updateUiState(events: List<Event>) {
    _uiState.update { currentState ->
        currentState.copy(
            upcomingEventList = EventList(
                allEvents = events,
                filteredEvents = events,
                selectedEvent = currentState.upcomingEventList.selectedEvent
            )
        )
    }
}
```

Which version would you say is easier to read/follow ?

app/src/main/java/com/github/se/eventradar/model/repository/event/IEventRepository.kt
```diff
@@ -15,4 +16,8 @@ interface IEventRepository {
   suspend fun deleteEvent(event: Event): Resource<Unit>
 
   suspend fun getEventsByIds(ids: List<String>): Resource<List<Event>>
+
+  fun observeEvents(): Flow<Resource<List<Event>>>
+
+  fun observeUpcomingEvents(userId: String): Flow<Resource<List<Event>>>
```
RaulinN, line 22: Yes this is great! The function signatures are exactly what I had in mind when discussing this solution last Friday! =)

app/src/main/java/com/github/se/eventradar/model/repository/event/FirebaseEventRepository.kt
```diff
@@ -80,4 +83,40 @@ class FirebaseEventRepository : IEventRepository {
...
+            return@addSnapshotListener
+          }
+          val events = snapshot?.documents?.mapNotNull { Event(it.data!!, it.id) }
+          trySend(Resource.Success(events ?: listOf()))
+        }
+    awaitClose { listener.remove() }
+  }
+
+  override fun observeUpcomingEvents(userId: String): Flow<Resource<List<Event>>> = callbackFlow {
+    val query = eventRef.whereArrayContains("attendees_list", userId)
```
RaulinN, line 102: Both your `observers` are good! Small detail for the future: you could cut the number of db calls in half. Observe upcoming events is in fact the list of all events filtered (i.e. dont need to receive the list twice)

You can see example number 1 and 2 in [this file](https://github.com/RaulinN/calculator-compose/blob/master/app/src/main/java/com/github/raulinn/calculator/CalculatorViewModel.kt) that I shared previously to answer the question about cognitive complexity

RaulinN (commented):
Good job!

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
Nath-Assa: Your suggestion is a lot easier to understand. Thank you!

app/src/main/java/com/github/se/eventradar/model/event/Event.kt
Nath-Assa: This code was introduced specifically for UI testing(I needed it to run the app) and was actually part of PR #270 , which was later merged. 
The issue, discussed with @Viccha18, arises when natural numbers are input into the database; they are recognized as Long types, we opted for this method since we didn't want values like 0 being cast to null. For latitude and longitude fields, we do not input natural numbers, so direct casting as Double suffices there. This convertToDouble function is a temporary measure and will be deprecated once the 'create Event' feature is implemented, ensuring all relevant values are explicitly managed as Double.

app/src/test/java/com/github/se/eventradar/EventOverviewViewModelTest.kt
```diff
@@ -441,4 +443,104 @@ class EventsOverviewViewModelTest {
...
+    mockkStatic(Log::class)
+    every { Log.d(any(), any()) } returns 0
+
+    (userRepository as MockUserRepository).updateCurrentUserId("user1")
+    val exception = Exception("Network error")
+    (eventRepository as MockEventRepository).eventsFlow.emit(Resource.Failure(exception))
+
+    EventsOverviewViewModel(eventRepository, userRepository)
+
+    delay(500)
```
jgeaso1266, line 539: nit: not sure exactly but these delay calls might not be good for regression (maybe at some point in the future it will take the `emit()` function 600 milliseconds to emit a failure instead of 500). Just a thought for future reference.

jgeaso1266 (approved):
one nit/warning for future reference, but otherwise LGTM

app/src/test/java/com/github/se/eventradar/EventOverviewViewModelTest.kt
Nath-Assa: It's something to keep in mind for future optimizations, as relying on fixed delays can introduce potential flakiness in tests if the underlying timings change.
 Note: It also worked with 100ms. I put 500 to be safe.

Nath-Assa:
Closes #252

PR #260, opened 2024-05-10
Fix Flaky Tests

jgeaso1266:
## Key Changes
- change function signature of signup and login tests
- create intents test rule to be used in both tests


jgeaso1266:
This is a lost cause: cannot figure out exact reason why certain tests are failing each time, and no other fixes will change that.

PR #268, opened 2024-05-13
Private Chat View Model

Nath-Assa:
Enhanced ChatViewModel.kt:
 - Added `private fun initOpponent` to set opponent details in UI state and `initAndGetMessages` to call `initOpponent` and fetch messages.


Nath-Assa:
Closing this PR, too many conflicts. This PR should only update ChatViewModel.kt and test it.

PR #269, opened 2024-05-14
Make Search Bar a Separate Reuasble Component

jgeaso1266:
## Objective
- refactor the Search Bar component of `SearchBarAndFilter` to be used in other components

## Key Changes
- [ ] put `SearchBarField` into separate composable function


christinehahaa (approved):
Tested on the machine and LGTM! Noted that coverage is below the threshold but once merged into main, will be okay.

PR #270, opened 2024-05-14
Buy ticket to Join Event

Viccha18:
# Objective

Write the code to allow a user to join an event.

# Details

- [x] Register the user in the attendee list of the event, register the event in the user attendingEventList. 
- [x] User can register to an event then view it in the upcoming list.
- [x] Ticket capacity is also decremented by one
- [x] Joining is a failure if no more ticket, or if any error happen
- [ ] Rollback in case of halfway error will be handled in the atomic version

I didn't implemented proper ticket selection since there is only one (so there is nothing to select, by default you buy the only existing ticket) and would take some time, I wanted to get the core functionality quickly done for MS3


app/src/androidTest/java/com/github/se/eventradar/event/EventSelectTicketUITest.kt
```diff
@@ -102,6 +111,51 @@ class EventSelectTicketUITest :
...
+  @Test
+  fun buyTicketFailure() = run {
+    ComposeScreen.onComposeScreen<EventSelectTicketScreen>(composeTestRule) {
+      buyButton {
+        assertIsDisplayed()
+        performClick()
+      }
+      // assert: the buy has been triggered
+      verify { mockViewModel.buyTicketForEvent() }
+      // confirmVerified(mockViewModel)
```
jgeaso1266, line 151: Why is this line commented out?

app/src/main/java/com/github/se/eventradar/ui/event/EventSelectTicket.kt
```diff
@@ -70,8 +104,8 @@ fun SelectTicket(viewModel: EventDetailsViewModel, navigationActions: Navigation
         // buy ticket button
         FloatingActionButton(
             onClick = {
-              context.toast("Buy ticket needs to be implemented")
-              /*TODO launch action when buying a ticket */
+              viewModel.buyTicketForEvent()
+              // context.toast("You registered to the event !")
```
jgeaso1266, line 108: Make sure to remove any commented code like this

app/src/main/java/com/github/se/eventradar/viewmodel/EventDetailsViewModel.kt
```diff
@@ -23,16 +28,33 @@ class EventDetailsViewModel
...
+          currentUserId = response.data
+        }
+        is Resource.Failure ->
+            Log.d(
+                "EventDetailsViewModel", "Could not get the user Id: ${response.throwable.message}")
+      }
+    }
   }
 
   private val _uiState = MutableStateFlow(EventUiState())
```
jgeaso1266, line 49: I did this in another PR but these variable declarations need to be at the _top_ of the function. They cannot go after the `init` function.

app/src/main/java/com/github/se/eventradar/viewmodel/EventDetailsViewModel.kt
```diff
@@ -61,6 +84,109 @@ constructor(
...
+  fun buyTicketForEvent() { // update the user attendee list, update the eventRepo ticket count
+    viewModelScope.launch {
+      if (!isTicketFree()) {
+        Log.d(
+            "EventDetailsViewModel",
+            "Paid tickets are not supported, all of them are considered free")
+      }
+
+      if (!isUserAttendingEvent()) {
+        // TODO needs to be atomic to avoid concurrency issues
```
jgeaso1266, line 96: I would not consider this a concern for the v1 (though it is a very good point)

app/src/main/java/com/github/se/eventradar/ui/event/EventDetails.kt
jgeaso1266, line 1: Currently, after registering for an event it goes back to the event details page. But that page does not automatically show that I have registered. It also still shows the FAB for me to register again, so it is not clear that the registration was successful. This can be fixed by a simple LaunchedEffect that checks if the user's registration status has been changed.

jgeaso1266 (changes requested):
One small bug and a few nits/questions.

app/src/main/java/com/github/se/eventradar/ui/event/EventDetails.kt
Viccha18: ooh, ok, thanks !

app/src/main/java/com/github/se/eventradar/viewmodel/EventDetailsViewModel.kt
jgeaso1266: Sorry I wasn't clear, I was talking about the TODO you put.

PR #273, opened 2024-05-15
Add New Ticket_Purchases Field

nikhen-s:
## Objective
-  Add a new ticket_purchases field to firebase

## Tasks done
- Edited ReadMe to explain the new ticket_purchases field
- Added field to firebase
- Added field to EventTickets and Refactored Tests Accordingly


Viccha18 (changes requested):
Thanks for the add !
It is just missing one line in `Event.kt` in function `fun toMap(): HashMap<String, Any> {...}` for `purchases` field

(Only a comment : Maybe its because of french but I feel like "purchased" suits better than "purchases" 🤔 )

nikhen-s:
> Thanks for the add ! It is just missing one line in `Event.kt` in function `fun toMap(): HashMap<String, Any> {...}` for `purchases` field
> 
> ```kotlin
> map["ticket_purchases"] = ticket.purchases
> ```
> 
> (Only a comment : Maybe its because of french but I feel like "purchased" suits better than "purchases" 🤔 )

I was thinking about this too haha :) But for me, ticket_purchased sounds more like a boolean field, whereas ticket_purchases sounds more like the number of ticket purchases. I hope it's not too confusing!

Viccha18 (approved):
LGTM, coverage is low but these changes do not add any new lines it only modify already existing one

PR #275, opened 2024-05-15
Hilt injection in tests + instantiation test for EventDetailsViewModel

Viccha18:
# Objective

Finish to set up hilt injection in tests (should be of use for E2E testing). As well as testing assisted-injected viewModels

Useful links from coach :
- https://developer.android.com/training/dependency-injection/hilt-testing?hl=fr
- https://www.youtube.com/watch?v=nOp_CEP_EjM


david-schroeter (commented):
Good job on this PR! This new way of testing to be able to directly use Hilt dependency injection will be very useful in a lot of places. Specially when you will write E2E tests!

I just wrote two suggestions, but the code overall was very well written, nice !

Issue #241, opened 2024-05-07
Private chat ViewModel


Viccha18:
View figma : private chat - individual chat

# Description

Implement the functionality of the private chat. 
- Typing a message
- Sending a message
- Filling the message list with new messages

Nath-Assa:
Closed in #282

PR #282, opened 2024-05-16
ChatViewModel Implementation + Tests

Nath-Assa:
### Key Changes:

- Implemented key functions in the ChatViewModel to enhance chat functionality and streamline operations
- Fixed `fun getMessages(user1: String, user2: String)` in `FirebaseMessageRepository.kt` (done by @christinehahaa)
- Updated scrollDown logic in `Chat.kt`. Scroll down should only be triggered when the current user sends a message, or loads the chat for the first time.
- Added comprehensive unit tests for all functions in the ViewModel to ensure reliability and functionality.


app/src/androidTest/java/com/github/se/eventradar/chat/ChatTest.kt
```diff
@@ -97,7 +97,7 @@ class ChatTest : TestCase(kaspressoBuilder = Kaspresso.Builder.withComposeSuppor
 
     mockUserRepository.updateCurrentUserId("1")
 
-    mockChatViewModel = ChatViewModel(mockMessageRepository, mockUserRepository)
+    mockChatViewModel = ChatViewModel(mockMessageRepository, mockUserRepository, "user1")
```
christinehahaa, line 101: technically should be the opponentId, so "2"

app/src/main/java/com/github/se/eventradar/ui/chat/Chat.kt
```diff
@@ -102,8 +100,8 @@ fun ChatScreenUi(
   val scrollState = rememberLazyListState(initialFirstVisibleItemIndex = messages.size)
   val messagesLoadedFirstTime = uiState.messagesLoadedFirstTime
   val messageInserted = uiState.messageInserted
-  LaunchedEffect(key1 = messagesLoadedFirstTime, messages, messageInserted) {
-    if (messages.isNotEmpty()) {
+  LaunchedEffect(key1 = messagesLoadedFirstTime, messageInserted) {
+    if ((messagesLoadedFirstTime || messageInserted) && messages.isNotEmpty()) {
```
christinehahaa, line 104: Good catch!

app/src/test/java/com/github/se/eventradar/ChatViewModelUnitTest.kt
```diff
@@ -0,0 +1,219 @@
...
+
+    val uiState = viewModel.uiState.value
+    val opponentResource = userRepository.getUser(opponentId)
+    assert(uiState.opponentProfile == nullUser)
+    verify {
+      Log.d(
+          "ChatViewModel",
+          "Error getting opponent details: ${(opponentResource as Resource.Failure).throwable.message}")
+    }
+    unmockkAll()
```
christinehahaa, line 165: Might be easier to unmock after every test in one go by adding this line at the start of the code, before all the tests. 

`  @After fun testTeardown() = runTest { unmockkAll() }
`

app/src/main/java/com/github/se/eventradar/viewmodel/ChatViewModel.kt
```diff
@@ -76,6 +171,7 @@ data class ChatUiState(
             user1ReadMostRecentMessage = false,
             user2ReadMostRecentMessage = false,
             messages = mutableListOf()),
+    val opponentId: String? = null, // To be deleted once VM is integrated with UI
```
christinehahaa, line 174: Is there a reason why it needs to be removed?

`opponentId` from the uiState is only used in the UI once in line 152. If it's not stored in the uiState, then the viewmodel needs to be passed in as a parameter so `viewModel.opponentId` instead of `uistate.opponentId`. 

The UiState is already passed in as a parameter so might be easiest to keep it in the UI. what do u think?

app/src/main/java/com/github/se/eventradar/ui/chat/Chat.kt
```diff
@@ -65,7 +62,8 @@ import java.time.LocalDateTime
 @Composable
 fun ChatScreen(viewModel: ChatViewModel = hiltViewModel(), navigationActions: NavigationActions) {
```
christinehahaa, line 63: I see you've made the factory changes in this PR. In that case, can you modify the UI at the same time?

This line should be:
`fun ChatScreen(viewModel: ChatViewModel, navigationActions: NavigationActions) {
`

I would suggest adding the nav already, but that can be done in the next PR, as we discussed :)

christinehahaa (commented):
This looks great Nathaniel, good job picking up on bugs and using factory in the VM. Just a few small comments + coverage needs to be increased slightly for CI to pass :)

app/src/test/java/com/github/se/eventradar/ChatViewModelUnitTest.kt
Nath-Assa: Since I don't use mockkStatic(...) in every test, I don't think it's necessary.

app/src/main/java/com/github/se/eventradar/ui/chat/Chat.kt
Nath-Assa: I believe this should also be done in the PR that updates Chat.kt to use the VM.

app/src/main/java/com/github/se/eventradar/viewmodel/ChatViewModel.kt
Nath-Assa: We should remove opponentId from ChatUiState to avoid redundancy. It's already accessible through viewModel.opponentId, making it unnecessary to track the same value in two places. This approach reduces complexity and potential errors. If adjusting the UI references from uiState.opponentId to viewModel.opponentId requires significant changes, we can evaluate the impact. However, streamlining this should generally be manageable.

christinehahaa: True, realised that now, thanks :)

christinehahaa (approved):
Great, LGTM now!

Nath-Assa:
Resolves #241

PR #290, opened 2024-05-16
Update Git Ignore

nikhen-s:
## Objective
- Updating gitignore to not include AndroidManifest.xml so we don't accidentally push our Map API Keys.


jgeaso1266:
This shouldn't be necessary. We already have secrets in gitignore and we need to track changes in Android Manifest as there are properties the all devs needs.

Issue #281, opened 2024-05-16
Private Chat UI bug fixes + Refine message repositories


christinehahaa:
Depends on Issue #280 

Fix bugs/issues from PR #291 

Acceptance Criteria:
- UI bugs described in PR #291 review are fixed
- Refined Message Repositories logic (Firebase + Mock) for `getMessages()` for 2 users, `addMessage()`, `createNewMessageHistory()`. `ChatViewModel.kt` may need an update depending on your new logic.
- Consequently, `FirebaseMessageRepositoryUnitTest.kt` `MockMessageRepositoryUnitTest.kt` and `ChatViewModelUnitTest.kt`, must also be revised to reflect these changes.

PR #291, opened 2024-05-17
Chat vm UI integration

christinehahaa:
Main changes:
- Private chat UI and VM connect
- Private chat screen accessible via the FAB in Messages nav bar > Friends tab > click to view friends profile
- Private chat screen accessible via the messages list (Messages nav bar > Messages tab)
- User can send messages, to someone they already have messaged or have never messaged
- If messaging someone they've never messaged before, new chats populate in the messages list (Messages nav bar > Messages tab)
- Lots of bug fixes to ensure the above functionality works 

To be completed:
- **When message bar is active, cannot see the top of the screen i.e. top bar is invisible and new messages sent are invisible unless you get rid of the keyboard and have the whole screen again.** 
- When on a private chat screen and navigating off the messages nav bar tab (e.g. to home) and then back to messages, view should reset to the messages list, but currently stays in the private chat screen. - NOT AN ISSUE
- When there's a lot of messages sent and they don't fit on the entire screen, cannot see the latest message sent unless you scroll down. Should scroll automatically to the latest message inserted. - FIXED BY JALEN
- Sometimes app signs you out when you click on messages in the messages list?? - NOT AN ISSUE
- **Need to link private chat screen to qr code scan - so when scanning QR code of a friend, goes to private chat screen (currently goes to home)**
- CI is failing due to buggy main (@nikhen-s  is fixing now) - DONE


app/src/main/java/com/github/se/eventradar/model/repository/message/FirebaseMessageRepository.kt
```diff
@@ -65,11 +66,12 @@ class FirebaseMessageRepository(db: FirebaseFirestore = Firebase.firestore) : IM
             .await()
 
     if (resultDocument == null || resultDocument.isEmpty) {
-      return createNewMessageHistory(user1, user2)
+      return Resource.Failure(Exception("No message history found between users"))
     }
 
     return try {
       val result = resultDocument.documents[0]
+      Log.d("FirebaseMessageRepository", "messages: $result")
```
Nath-Assa, line 74: This can be removed.

app/src/main/java/com/github/se/eventradar/model/repository/message/FirebaseMessageRepository.kt
```diff
@@ -65,11 +66,12 @@ class FirebaseMessageRepository(db: FirebaseFirestore = Firebase.firestore) : IM
             .await()
 
     if (resultDocument == null || resultDocument.isEmpty) {
-      return createNewMessageHistory(user1, user2)
+      return Resource.Failure(Exception("No message history found between users"))
```
Nath-Assa, line 69: Question: I believe what we had before worked fine and now there is an inconsistency with `getMessages(uid: String)`. Can you explain your reasoning?

app/src/main/java/com/github/se/eventradar/model/repository/message/FirebaseMessageRepository.kt
```diff
@@ -87,7 +89,7 @@ class FirebaseMessageRepository(db: FirebaseFirestore = Firebase.firestore) : IM
       val messageHistory = MessageHistory(messageHistoryMap, result.id)
       Resource.Success(messageHistory)
     } catch (e: Exception) {
-      createNewMessageHistory(user1, user2)
+      Resource.Failure(e)
```
Nath-Assa, line 92: Nice catch! Improves consistency with `getMessages(uid: String)`

app/src/main/java/com/github/se/eventradar/ui/chat/Chat.kt
```diff
@@ -61,24 +59,19 @@ import com.github.se.eventradar.viewmodel.ChatViewModel
...
+fun ChatScreen(viewModel: ChatViewModel, navigationActions: NavigationActions) {
   val uiState by viewModel.uiState.collectAsState()
-  // Note the opponentId should no longer be in the uiState but be accessible with
-  // viewModel.opponentId
-  // TO DO: Implement get messages between two users in VM
-  //    viewModel.getMessages(senderId, opponentId)
 
-  val context = LocalContext.current // only needed until message send is implemented
+  viewModel.initOpponent()
+  viewModel.getMessages()
```
Nath-Assa, line 66: The functions above are already called when the viewModel is created. View ChatViewModel's `init` block.

app/src/main/java/com/github/se/eventradar/ui/chat/Chat.kt
```diff
@@ -264,7 +250,6 @@ fun ChatScreenPreview() {
   val sampleUiState =
```
Nath-Assa, line 250: Suggestion: The ChatScreenPreview() can be updated.
View this example from `Home.kt`:
```
@Preview(showBackground = true, showSystemUi = true)
@ExcludeFromJacocoGeneratedReport
@Composable
fun HomeScreenPreview() {
  val mockEventRepo = MockEventRepository()
  val mockUserRepo = MockUserRepository()
  HomeScreen(
      EventsOverviewViewModel(mockEventRepo, mockUserRepo),
      NavigationActions(rememberNavController()))
}
```

app/src/main/java/com/github/se/eventradar/model/repository/message/FirebaseMessageRepository.kt
Nath-Assa: After discussing with @christinehahaa , the reasoning is that we shouldn't create a history for users that have not yet texted each other.

app/src/main/java/com/github/se/eventradar/ui/chat/Chat.kt
```diff
@@ -117,8 +106,6 @@ fun ChatScreenUi(
     }
```
Nath-Assa, line 106: **UI bug fix:**
The `LaunchedEffect(key1 = imeBottomPadding)` above is the reason why when we start typing a message and the keyboard opens, all previous messages disappear. Consider removing lines 101-106.

Nath-Assa (changes requested):
Overall, good work. Some tests still need to be updated and I made a few suggestions.

app/src/main/java/com/github/se/eventradar/model/message/MessageHistory.kt
```diff
@@ -29,7 +29,6 @@ data class MessageHistory(
     map["latest_message_id"] = latestMessageId
     map["from_user_read"] = user1ReadMostRecentMessage
     map["to_user_read"] = user2ReadMostRecentMessage
-    map["messages"] = messages.map { it.toMap() }
```
Nath-Assa, line 32: This change causes a test in `FirebaseMessageRepositoryUnitTest.kt` to fail.
What is your reasoning?

app/src/main/java/com/github/se/eventradar/model/repository/message/FirebaseMessageRepository.kt
Nath-Assa, line 1: You've updated the logic in FirebaseMessageRepository.kt. Consequently, this requires corresponding updates in FirebaseMessageRepositoryUnitTest.kt and MockMessageRepository.kt to ensure alignment with the new logic. Additionally, the MockMessageRepositoryUnitTest.kt and ChatViewModelUnitTest.kt, which depend on MockMessageRepository.kt, must also be revised to reflect these changes.

Nath-Assa:
UI tests need to be updated.

app/src/main/java/com/github/se/eventradar/model/message/MessageHistory.kt
jgeaso1266: I don't think that messages should be updated when converting message history to map. It depends on how it is done on the firebase side, but if messages are automatically added to firebase when adding messages, then this is not needed.

app/src/main/java/com/github/se/eventradar/model/repository/message/FirebaseMessageRepository.kt
```diff
@@ -96,9 +98,25 @@ class FirebaseMessageRepository(db: FirebaseFirestore = Firebase.firestore) : IM
       messageHistory: MessageHistory
```
Nath-Assa, line 98: Suggested Update:

In the current implementation of FirebaseMessageRepository.kt, the method addMessage() requires an input of an empty MessageHistory when two users have not previously exchanged messages. This method then does not update as expected because it creates a new MessageHistory. To improve this, it would be more functional to modify the method to take in 2 uids, then start by calling getMessages(user1, user2) at the beginning of the interaction. This change would ensure that if a message history between the two users doesn't exist, it is created appropriately, and if it does exist, it is retrieved and updated correctly.

app/src/main/java/com/github/se/eventradar/model/message/MessageHistory.kt
Nath-Assa: Actually this is now handled in **addMessage()**. I kept this change and fixed the failing test.

app/src/main/java/com/github/se/eventradar/model/repository/message/FirebaseMessageRepository.kt
christinehahaa: Agree, I was afraid to change the function signature since Jalen created it and it had dependencies on tests etc. But this is the ideal structure.

app/src/androidTest/java/com/github/se/eventradar/messages/MessagesTest.kt
```diff
@@ -70,17 +70,21 @@ class MessagesTest : TestCase(kaspressoBuilder = Kaspresso.Builder.withComposeSu
               username = "Test$i"))
 
       // Add messages to the mock repository (except for the first user)
+      // TODO: fix logic with add message
```
jgeaso1266, line 73: This will be fixed in #281

app/src/main/java/com/github/se/eventradar/model/repository/message/FirebaseMessageRepository.kt
jgeaso1266: The easiest solution is to have `addMessage` behave as it did before (precondition is that a message history is already created and in the repo), and then only check if messages are in message history on the messages overview screen (if there aren't any messages then don't show the message history). This will be implemented in #281.

app/src/test/java/com/github/se/eventradar/ChatViewModelUnitTest.kt
```diff
@@ -167,6 +157,50 @@ class ChatViewModelUnitTest {
...
+    (userRepository as MockUserRepository).updateCurrentUserId("user1")
+    userRepository.addUser(opponent)
+
+    val msg1 = mockMessage.copy(sender = "user1", id = "msg1")
+    val msg2 = mockMessage.copy(sender = "user2", id = "msg2")
+
+    // TODO: clean up logic with add message
+    val messageHistory = messageRepository.createNewMessageHistory("user1", "user2")
+
+    val editedMH = (messageHistory as Resource.Success).data
```
jgeaso1266, line 171: tests will also be changed to match with above fix in #281.

jgeaso1266 (approved):
Some changes will be made in #281 but great start! LGTM

Nath-Assa (commented):
This pull request introduces important features and enhancements. While the core functionality meets the requirements for milestone MS3, there are still some UI bugs that need addressing, and the code, particularly in the Message Repositories, could benefit from further optimization. These issues do not impede the current functionality but could impact the user experience. Therefore, I recommend merging this pull request as part of MS3, and we can prioritize these improvements in the next sprint to refine the experience and maintainability of our code.

wisdomseeker2001:
Overall, not the best person to review but I did what I could. again I would like to review it properly by testing it on emulator via preview etc but build fails.

Nath-Assa:
All issues will be fixed in https://github.com/swent-sp-2024-event-radar/event-radar-app/issues/281

PR #293, opened 2024-05-17
Use scaffold layout instead of constraint layout

### LINKED ISSUE #264: Use Scaffold instead of Constraint Layout
## Objective
- remove the boilerplate associated with using a constraint layout and use scaffold instead

## Acceptance Criteria
- all constraint layouts are replaced with scaffolds

jgeaso1266:
## Objective
- remove the boilerplate associated with using a constraint layout and use scaffold instead

## Key Changes
- [ ] use scaffold layout in Event Details instead of constraint layout
- [ ] reformat Event Details to match the Figma
- [ ] surround constraint layout in Home with scaffold layout and move bottom and top bars into the scaffold constructor
- [ ] update tests where needed

## Demo

closes #264

### DIFF
--- a/app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
+++ b/app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
@@ -9,6 +9,7 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
 import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Scaffold
 import androidx.compose.material3.Tab
 import androidx.compose.material3.TabRow
 import androidx.compose.material3.Text
@@ -56,6 +57,7 @@ fun HomeScreen(
 ) {
   // Ui States handled by viewModel
   val uiState by viewModel.uiState.collectAsState()
+
   LaunchedEffect(key1 = uiState.isSearchActive, key2 = uiState.isFilterActive) {
     if (uiState.isSearchActive || uiState.isFilterActive) {
       viewModel.filterEvents()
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/github/se/eventradar/ui/home/Home.kt
```diff
@@ -69,247 +71,247 @@ fun HomeScreen(
...
+                  painter = painterResource(id = R.drawable.event_logo),
+                  contentDescription = "Event Radar Logo",
+                  modifier = Modifier.size(width = 186.dp, height = 50.dp))
+            }
+      },
+      bottomBar = {
+        BottomNavigationMenu(
+            onTabSelected = navigationActions::navigateTo,
+            tabList = TOP_LEVEL_DESTINATIONS,
+            selectedItem = getTopLevelDestination(Route.HOME),
```
Viccha18, line 90: Would be nice to use that notation `selectedItem = getTopLevelDestination(Route.HOME)` in EventDetails & EventSelectTicket too ! (It was my mistake when I've created those files but if you could change it, I would be grateful 🙏 )

app/src/main/res/values/strings_event.xml
```diff
@@ -4,6 +4,8 @@
     <string name="event_location">Location</string>
     <string name="event_distance">Distance</string>
     <string name="event_date_and_time">Date and time</string>
+    <string name="event_date">Date</string>
```
Viccha18, line 7: I think we can remove this if it is replaced by `<string name="event_date_and_time">Date and time</string>`
```
<string name="event_date">Date</string>
<string name="event_time">Time</string>
```

Viccha18 (commented):
Thank you for the changes ! The code is good👍 and it looks nice on the home screen ! However, for the event details, I would suggest increasing the padding in general to be more in accordance with how the figma looks. Here is it how it looks on the emulator vs figma : 
image

especially the left and right border as well as the title which miss some space with the image

Viccha18 (commented):
Good addition for modularity ! 👍 I wasn't the one who worked on Home.kt but it looks more clean this way. `getEventsBasedOffUiState()` was a nice addition as well as `EventOverview`. Exctracting the tabs and search & filter could be nice too for the lisibility of the HomeScreen()

Viccha18 (approved):
LGTM thank you !

PR #300, opened 2024-05-19
Scan friend improvements

wisdomseeker2001:
Hi:

I have attempted to make improvements based on coaches feedback. The Viewmodel should not be making in any asynchronous calls within its init block. Reason being is VM can be retruned without the aysnc calls being completed and hence VM can be used when in a non complete state which can cause race conditions. 

I have hence redesigned the VM, it will be complicated to understand because I have used notions no one has explored yet:

I used improvement 1 of https://proandroiddev.com/mastering-android-viewmodels-essential-dos-and-donts-part-1-%EF%B8%8F-bdf05287bca9 - this is the only way you can understand what and why i did it this way.

another chnage I made the switching of screen code upon scanning a qr appear in the VM instead of in the UI. before i would change the state in the VM and then the UI would recat to this state and chnage screen. this is not good as can cause race conditions. for more context: https://discord.com/channels/1215608534088024145/1216071194915246102/1241442265411354635 is the coaches feedback on the issues I am/ was facing. 

I also changed to switching of the screen to the chat screen ui now merged into main by Christine (line 221 of VM file). 

everything passes locally and on the CI. unfortunately I have 79.1% code coverage and dont really know what to increase it.

### DIFF
--- a/app/src/main/java/com/github/se/eventradar/ui/navigation/NavGraph.kt
+++ b/app/src/main/java/com/github/se/eventradar/ui/navigation/NavGraph.kt
@@ -21,7 +21,6 @@ import com.github.se.eventradar.util.toast
 import com.github.se.eventradar.viewmodel.ChatViewModel
 import com.github.se.eventradar.viewmodel.EventDetailsViewModel
 import com.github.se.eventradar.viewmodel.ViewFriendsProfileViewModel
-import com.github.se.eventradar.viewmodel.qrCode.ScanFriendQrViewModel
 
 @Composable
 fun NavGraph(
@@ -64,10 +63,7 @@ fun NavGraph(
         }
     // TODO replace the Toast message with the corresponding screen function of the route
     composable(Route.MESSAGE) { MessagesScreen(navigationActions = navActions) }
-    composable(Route.SCANNER) {
-      val viewModel = ScanFriendQrViewModel.create(navigationActions = navActions)
-      QrCodeScreen(viewModel = viewModel, navigationActions = navActions)
-    }
+    composable(Route.SCANNER) { QrCodeScreen(navigationActions = navActions) }
     composable(Route.PROFILE) {
       HomeScreen(navigationActions = navActions)
       context.toast("Profile screen needs to be implemented")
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanFriendQrViewModel.kt
```diff
@@ -121,16 +211,18 @@ constructor(
     return updateResult is Resource.Success
   }
 
-  fun resetNavigationEvent() {
-    _uiState.value = _uiState.value.copy(action = Action.None)
+  private fun changeAction(action: Action) {
+    _uiState.value = uiState.value.copy(action = action)
+    if (action == Action.NavigateToNextScreen) {
+      navigationActions.navController.navigate(
```
jgeaso1266, line 217: any calls to the navigation actions should be in the UI state. This is how all the other screens operate and it means you don't have to worry about assisted hilt injection with the navigation actions

app/src/main/java/com/github/se/eventradar/ui/navigation/NavGraph.kt
```diff
@@ -61,10 +62,12 @@ fun NavGraph(
           val viewModel = ViewFriendsProfileViewModel.create(friendUserId = friendUserId)
           ViewFriendsProfileUi(viewModel = viewModel, navigationActions = navActions)
         }
-
     // TODO replace the Toast message with the corresponding screen function of the route
     composable(Route.MESSAGE) { MessagesScreen(navigationActions = navActions) }
-    composable(Route.SCANNER) { QrCodeScreen(navigationActions = navActions) }
+    composable(Route.SCANNER) {
+      val viewModel = ScanFriendQrViewModel.create(navigationActions = navActions)
+      QrCodeScreen(viewModel = viewModel, navigationActions = navActions)
```
jgeaso1266, line 69: There shouldn't be a need to pass navigation actions into both the UI and the VM.

jgeaso1266 (commented):
A few comments. I read the article you linked in the description and it helped in understanding your changes, but I am not sure why you are passing in the nav actions to the view model. I see that you use it in the VM, but again, I don't see why this can't still be handled in the UI (this is how all the other screens are done as well by the way). Also there are a lot of tests you commented out. Will these be edited and added back in? If not, then all commented code should be removed.

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanFriendQrViewModel.kt
wisdomseeker2001: I made these changes because the coach told me to do so, read discoord to understand why it was causing race conditions

PR #301, opened 2024-05-20
Add registration cancellation method in Event VM

jgeaso1266:
## Objective
- Implement VM for Cancel Registration user flow

## Key Changes
- [ ] add a method to remove a user from an event
- [ ] add unit tests

### DIFF
--- a/app/src/main/java/com/github/se/eventradar/viewmodel/EventDetailsViewModel.kt
+++ b/app/src/main/java/com/github/se/eventradar/viewmodel/EventDetailsViewModel.kt
@@ -245,9 +245,9 @@ constructor(
             Log.d("EventDetailsViewModel", "Error getting user: ${userResponse.throwable.message}")
           }
         }
+
+        getEventData()
       }
-      
-      getEventData()
     } else {
       errorOccurred.value = true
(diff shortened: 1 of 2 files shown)


Viccha18 (commented):
LGTM, just make sure to change the log.d message I've pointed out in EventDetailsViewModel.

Also, just a note : I think that function will need to be an atomic transaction too

app/src/main/java/com/github/se/eventradar/viewmodel/EventDetailsViewModel.kt
```diff
@@ -190,6 +190,67 @@ constructor(
...
+          }
+          is Resource.Failure -> {
+            errorOccurred.value = true
+            Log.d("EventDetailsViewModel", "Error getting user: ${userResponse.throwable.message}")
+          }
+        }
+      }
+    } else {
+      errorOccurred.value = true
+      Log.d("EventDetailsViewModel", "Error updating event data: No more tickets!}")
```
jgeaso1266, line 250: the message should be "no event found"

Viccha18 (approved):
LGTM

PR #304, opened 2024-05-21
My Hostings - search & filter

christinehahaa:
Main changes:
- added UI + VM for search and filter in my hostings page

To be completed:
- Convert "my hostings" page to scaffold > constraint layout

### DIFF
--- a/app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
+++ b/app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
@@ -59,7 +59,7 @@ fun HostingScreen(
   val uiState by viewModel.uiState.collectAsState()
   LaunchedEffect(key1 = uiState.isSearchActive, key2 = uiState.isFilterActive) {
     if (uiState.isSearchActive || uiState.isFilterActive) {
-      viewModel.filterEvents()
+      viewModel.filterHostedEvents()
     }
   }
   LaunchedEffect(Unit) { viewModel.getHostedEvents() }
--- a/app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
+++ b/app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
@@ -33,9 +33,12 @@ import androidx.compose.ui.unit.sp
 import androidx.constraintlayout.compose.ConstraintLayout
 import androidx.hilt.navigation.compose.hiltViewModel
 import com.github.se.eventradar.R
+import com.github.se.eventradar.model.event.Event
 import com.github.se.eventradar.ui.BottomNavigationMenu
 import com.github.se.eventradar.ui.component.EventList
+import com.github.se.eventradar.ui.component.FilterPopUp
 import com.github.se.eventradar.ui.component.Logo
+import com.github.se.eventradar.ui.component.SearchBarAndFilter
 import com.github.se.eventradar.ui.component.ViewToggleFab
 import com.github.se.eventradar.ui.component.getIconFromViewListBool
 import com.github.se.eventradar.ui.map.EventMap
@@ -44,6 +47,7 @@ import com.github.se.eventradar.ui.navigation.Route
 import com.github.se.eventradar.ui.navigation.TOP_LEVEL_DESTINATIONS
 import com.github.se.eventradar.ui.navigation.getTopLevelDestination
 import com.github.se.eventradar.util.toast
+import com.github.se.eventradar.viewmodel.HostedEventsUiState
 import com.github.se.eventradar.viewmodel.HostedEventsViewModel
 
 @Composable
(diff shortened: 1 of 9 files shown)


wisdomseeker2001:
Overall, the testing is super clear and nicely written, will definetely inspire myself from how you wrote them, code coverage is hence solid. 

All Ui components you added and VM functions look clean and use common practices and notions we have all used. I do think it is worth thinking of what some logic being re assigned to VM/ UI based on some of my comments.  Code duplication is a big problem jusy because a lot is in commong with EventDetailsOverView so you will need to find a way to share the code between both (sonarcloud precises which block of code)

christinehahaa:
Regarding code duplication - I have looked into ways to reduce it. It will require some major code refactoring, which I don't think is worth the time. 

Code duplication after merge is estimated to be 1.4%. I think this is acceptable. We can always add "Remove code duplication" as a task in the SB and see if we have time for it.

app/src/main/java/com/github/se/eventradar/viewmodel/SearchFilterUiState.kt
```diff
@@ -0,0 +1,15 @@
+package com.github.se.eventradar.viewmodel
+
+import com.github.se.eventradar.model.event.EventCategory
+
+// This class represents a common uiState for search and filter operations
+// It allows uiState in FilterPopUp to be of either type EventsOverviewUiState or HostedEventUiState
+sealed class SearchFilterUiState {
```
nikhen-s, line 7: 👍

nikhen-s (changes requested):
Looks good & works for me!

Main comments:

- A very small nit (but I don't think we need to change it). Maybe it's just a me thing but I thought that we could press on the search trailing icon on the right side of the search bar (but I understand that this is not intended behavior). Users (Coaches) might think it's a bug if they can't press on the search icon, so maybe we should remove it? But lmk!

- The Code duplication is mostly because we don't have a search bar and filter viewModel - was it more difficult to implement than expected, if so I think it's no worries that there is some code duplication as we need to focus on important tasks, but just wanted to know your reasoning.

- Finally I left just a small question for the radius part of the filter function.

That's all & great job!

app/src/main/java/com/github/se/eventradar/viewmodel/SearchFilterUiState.kt
nikhen-s, line 7: works for me! this keeps the state in the search bar and filter upon switching tabs in the app, which i think may not be the same if it's a normal data class 👍

christinehahaa:
> * The Code duplication is mostly because we don't have a search bar and filter viewModel, instead we have a UI State - was it more difficult to implement than expected, if so I think it's no worries that there is some code duplication as we need to focus on important tasks, but just wanted to know your reasoning.
> *

It is not necessarily difficult, but will just take time (~2-3hours + potential changes to tests etc.). Will need to extract the search and filter functions + potentially pass them into the existing viewmodels if there's dependencies or remove these dependencies. Duplication after merge is estimated 1.4%. 

I would also prefer to work on another feature, since we have a few to implement still before M4.

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -118,6 +139,25 @@ fun HostingScreen(
...
               iconVector = getIconFromViewListBool(uiState.viewList))
         }
+
+    if (uiState.isFilterDialogOpen) {
+      FilterPopUp(
+          onFreeSwitchChanged = { viewModel.onFreeSwitchChanged() },
+          onFilterApply = {
+            if (uiState.radiusQuery != "") {
+              viewModel.onRadiusQueryChanged("")
+              context.toast("Radius filtering not yet implemented")
```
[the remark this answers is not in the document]
christinehahaa: I think Jalen actually added this when he refactored the code to scaffold > constraint layout - I've asked him why.

nikhen-s:
> > * A very small nit (but I don't think we need to change it). Maybe it's just a me thing but I thought that we could press on the search trailing icon on the right side of the search bar (but I understand that this is not intended behavior). Users (Coaches) might think it's a bug if they can't press on the search icon, so maybe we should remove it? But lmk!
> 
> Good point. I like that there's an icon to symbolise searching but agree it can confuse the user if there's no onClick. Does making it a leading icon make it clearer that it's not clickable? If so, I can change it to a leading icon.
> 
> > * The Code duplication is mostly because we don't have a search bar and filter viewModel, instead we have a UI State - was it more difficult to implement than expected, if so I think it's no worries that there is some code duplication as we need to focus on important tasks, but just wanted to know your reasoning.
> 
> It is not necessarily difficult, but will just take time (~2-3hours + potential changes to tests etc.). Will need to extract the search and filter functions + potentially pass them into the existing viewmodels if there's dependencies or remove these dependencies. Duplication after merge is estimated 1.4%.
> 
> I would also prefer to work on another feature, since we have a few to implement still before M4.

Regarding the code duplication, I am ok with it and we have other things to prioritise. About the search bar icon, I think moving it to the left could help it seem more as a decorative than something to click on :)

nikhen-s (approved):
LGTM + making the search icon a leading one is great 👍

PR #307, opened 2024-05-22
Private chat screen - bug fixes + refine messages repo

christinehahaa:
Main changes:
- Fixed the bloody nasty bug where top app bar wasn't showing when keyboard typing in message bar by:
   - Changing app visibility when keyboard is open to squish screen size and display all elements 
   - This meant creating a new state and only displaying the bottom nav bar if they keyboard was not open
- Refined Message Repositories logic (Firebase + Mock) for `getMessages()` for 2 users, `addMessage()`, `createNewMessageHistory()`
- Changed associated tests in `FirebaseMessageRepositoryUnitTest.kt` `MockMessageRepositoryUnitTest.kt` and `ChatViewModelUnitTest.kt`

### DIFF
--- a/app/src/test/java/com/github/se/eventradar/ChatViewModelUnitTest.kt
+++ b/app/src/test/java/com/github/se/eventradar/ChatViewModelUnitTest.kt
@@ -313,12 +313,13 @@ class ChatViewModelUnitTest {
 
     val expectedMH =
         MessageHistory(
-            user1 = "Default sender",
-            user2 = "Default recipient",
-            latestMessageId = "DefaultId",
+            user1 = "user1",
+            user2 = "user2",
+            latestMessageId = "",
             user1ReadMostRecentMessage = false,
             user2ReadMostRecentMessage = false,
-            messages = mutableListOf())
+            messages = mutableListOf(),
+            id = "0")
 
     verify { Log.d("ChatViewModel", expectedLogMessage) }
     assert(viewModel.uiState.value.messageHistory == expectedMH)
--- a/app/src/test/java/com/github/se/eventradar/ChatViewModelUnitTest.kt
+++ b/app/src/test/java/com/github/se/eventradar/ChatViewModelUnitTest.kt
@@ -18,7 +18,6 @@ import java.time.LocalDateTime
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.delay
-import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.test.TestDispatcher
 import kotlinx.coroutines.test.UnconfinedTestDispatcher
 import kotlinx.coroutines.test.resetMain
@@ -175,8 +174,6 @@ class ChatViewModelUnitTest {
 
     viewModel =
         ChatViewModel(messageRepository, userRepository, opponentId) // Initialize view model
-
-    runBlocking { viewModel.getMessages() }
 
     val expectedMessages = mutableListOf(msg1, msg2)
     val uiState = viewModel.uiState.value
--- a/app/src/test/java/com/github/se/eventradar/ChatViewModelUnitTest.kt
+++ b/app/src/test/java/com/github/se/eventradar/ChatViewModelUnitTest.kt
@@ -167,13 +167,11 @@ class ChatViewModelUnitTest {
     val msg1 = mockMessage.copy(sender = "user1", id = "msg1")
     val msg2 = mockMessage.copy(sender = "user2", id = "msg2")
 
-    val messageHistory = messageRepository.createNewMessageHistory("user1", "user2")
-    val editedMH = (messageHistory as Resource.Success).data
+    val resource = messageRepository.createNewMessageHistory("user1", "user2")
+    val messageHistory = (resource as Resource.Success).data
 
-      messageRepository.addMessage(msg1, editedMH)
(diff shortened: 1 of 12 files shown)


christinehahaa:
> After testing the app, it appears to function nearly perfectly. However, I've identified a problem. Specifically, when a conversation exceeds one page in length, the keyboard obscures the most recent messages(view images below). Ideally, the latest message should remain visible even when the keyboard is active.

Nice pick up, I've fixed the bug now :)

app/src/test/java/com/github/se/eventradar/ChatViewModelUnitTest.kt
```diff
@@ -299,6 +299,25 @@ class ChatViewModelUnitTest {
     assert(viewModel.uiState.value.messageHistory == (expected as Resource.Success).data)
   }
 
+  @Test
+  fun `observeMessages Success No Existing Message History`() = runTest {
```
Nath-Assa, line 303: Suggestion: Consider deleting this test which seems to be extremely similar to : `observeMessages Success`

app/src/test/java/com/github/se/eventradar/ChatViewModelUnitTest.kt
Nath-Assa, line 1: Nit: Although I can't comment directly on the lines as they haven't been added or removed, many tests in this file need to be revised for consistency with the updated logic. Previously, the `addMessage()` function would accept a MessageHistory with an empty message list and create a new MessageHistory, effectively making the inputted MessageHistory irrelevant. This approach is still reflected in the tests. For example:
`val messageHistory = messageRepository.createNewMessageHistory("user1", "user2")
val editedMH = (messageHistory as Resource.Success).data
editedMH.messages.add(msg1)`

This should be updated to:
`val messageHistory = messageRepository.createNewMessageHistory("user1", "user2")
val editedMH = (messageHistory as Resource.Success).data
messageRepository.addMessage(msg1, editedMH)`

app/src/test/java/com/github/se/eventradar/FirebaseMessageRepositoryUnitTest.kt
Nath-Assa, line 1: Nit: The function test `addMessage() with empty message history` needs an update to reflect recent changes. Previously, addMessage() would create a new MessageHistory when the message list was empty. However, this behavior has been modified, and the test should be adjusted accordingly.

Nath-Assa (commented):
Great work! I have tested the app and it works perfectly. 
**Suggestion:**
Update MockMessageRepository and MockMessageRepositoryUnitTest to mimic the latest version of FirebaseMessageRepository(use version control). After these adjustments, please update the ChatViewModelUnitTest to align with the MockMessageRepository. 

I have also added other suggestions in my comments.

christinehahaa:
Thanks for all comments Nathaniel. 

I have updated tests as per comments. I have one failing test `observeMessages Failure` in the `ChatViewModelUnitTest.kt` which I am unable to solve. If you could assist and provide insights, as you created the test, that would be great.

Nath-Assa:
> Thanks for all comments Nathaniel.
> 
> I have updated tests as per comments. I have one failing test `observeMessages Failure` in the `ChatViewModelUnitTest.kt` which I am unable to solve. If you could assist and provide insights, as you created the test, that would be great.

After debugging the test, I discovered the cause of the failure. The viewModel no longer generates an empty message history when a failure occurs. Additionally, during initialization, invoking `getMessages` triggers `getMessages(user1, user2)`, creating a new message history for the two users. However, the `expectedMH` in our test did not match the output from `createMessageHistory`, leading to the discrepancy.

app/src/test/java/com/github/se/eventradar/MockMessageRepositoryUnitTest.kt
```diff
@@ -29,77 +29,70 @@ class MockMessageRepositoryUnitTest {
...
     var messageHistory = messageRepository.getMessages("1", "2")
 
-    assert(messageHistory is Resource.Failure)
+    assert(messageHistory is Resource.Success)
 
-    messageHistory = messageHistory as Resource.Failure
-    assert(messageHistory.throwable.message == "No message history found between users")
+    messageHistory = messageHistory as Resource.Success
+    assert(messageHistory.data.user1 == "1")
+    assert(messageHistory.data.user2 == "2")
```
Nath-Assa, line 40: Nit: here we could add additional assertions for the other fields.

app/src/test/java/com/github/se/eventradar/ChatViewModelUnitTest.kt
```diff
@@ -167,14 +167,11 @@ class ChatViewModelUnitTest {
...
-    val editedMH = (messageHistory as Resource.Success).data
-
-    editedMH.messages.add(msg1)
-
-    messageRepository.addMessage(msg2, editedMH)
+    messageRepository.addMessage(msg1, messageHistory)
+    messageRepository.addMessage(msg2, messageHistory)
 
     viewModel =
         ChatViewModel(messageRepository, userRepository, opponentId) // Initialize view model
```
Nath-Assa, line 177: Nit: the line below: ` runBlocking { viewModel.getMessages() }` is not necessary as it already called upon init.

Nath-Assa (approved):
Great work! just a few nits.

christinehahaa:
Regarding the 1 failing test, thanks for fixing it up. I had made those changes but didn't include the id. Also made changes to the nits, so now LGTM.

PR #310, opened 2024-05-23
Change signup DOB separator from "/" to "." 

Viccha18:
# Objective

Reintroduce nummeric keyboard to type date of birth in the signup screen. The "/" wasn't available to type with the nummeric keyboard so it is changed to "." which also makes it more consistent with the how the dates are displayed in the rest of the app.


wisdomseeker2001:
LGTM, nothing I can see to improve ! Great Work

wisdomseeker2001 (approved):

PR #311, opened 2024-05-23
Some atomicity for join/leave event

Viccha18:
# Objective

Use some atomic updates function in the join / leave event functions to avoid race condition and one user cancelling the registration of another.

Useful links :
- https://firebase.google.com/docs/firestore/manage-data/add-data
- https://firebase.google.com/docs/firestore/manage-data/transactions

This will also fix the bug of ticket purchases not being updated

# Note

This current code is not perfect. There are 3 updates needed to fulfill a (un)registration to an event
- update the event attendee list
- (in/de)crement the ticket purchases
- update the user attendance list

Currently, these three operations are atomic but don't happen together as a whole. If one fail, the others won't rollback.


app/src/main/java/com/github/se/eventradar/model/repository/event/FirebaseEventRepository.kt
```diff
@@ -118,4 +120,65 @@ class FirebaseEventRepository(db: FirebaseFirestore = Firebase.firestore) : IEve
...
+    }
+  }
+
+  override suspend fun decrementPurchases(eventId: String): Resource<Unit> {
+    return addSubPurchases(eventId, -1)
+  }
+
+  private fun addSubPurchases(eventId: String, value: Long): Resource<Unit> {
+    return try {
+      eventRef.document(eventId).update("ticket_purchases", FieldValue.increment(value))
```
nikhen-s, line 178: maybe it's also useful to add a check here to ensure that the ticket_purchases value does not become less than zero and throw an error if it does :)

app/src/main/java/com/github/se/eventradar/model/repository/event/MockEventRepository.kt
```diff
@@ -75,4 +76,64 @@ class MockEventRepository : IEventRepository {
       }
     }
   }
+
+  override suspend fun addAttendee(eventId: String, attendeeUserId: String): Resource<Unit> {
+    return when (val res = getEvent(eventId)) {
+      is Resource.Success -> {
+        res.data?.attendeeList?.add(attendeeUserId)
```
nikhen-s, line 83: question: this has to be checked for null because getEvent returns Resource<Event?>, what do you think about changing the return type of getEvent to Resource<Event> (in both Mock and Firebase Repo) and doing the null checks for the data in getEvent, so that we don't have to use the `?` in `res.data?.attendeeList?.add(attendeeUserId)`? Maybe this is too much refactoring though but helps reduce the possibility of null event data being used in other functions

app/src/main/java/com/github/se/eventradar/model/repository/event/FirebaseEventRepository.kt
```diff
@@ -118,4 +120,65 @@ class FirebaseEventRepository(db: FirebaseFirestore = Firebase.firestore) : IEve
...
+        }
+      }
+      return Resource.Success(Unit)
+    } catch (e: Exception) {
+      return Resource.Failure(e)
+    }
+  }
+
+  override suspend fun decrementPurchases(eventId: String): Resource<Unit> {
+    return addSubPurchases(eventId, -1)
```
nikhen-s, line 173: question: was there a reason you separated the function into `addSubPurchases` and `decrementPurchases` instead of including it as one?

app/src/main/java/com/github/se/eventradar/viewmodel/EventDetailsViewModel.kt
```diff
@@ -251,7 +275,60 @@ constructor(
     } else {
       errorOccurred.value = true
       Log.d("EventDetailsViewModel", "Error getting event data: no event displayed")
+    }*/
+
+  }
+
+  private suspend fun unregistrationProcedure(eventId: String, userId: String): Boolean {
```
nikhen-s, line 282: nice functions to help decompose the task :)

nikhen-s (changes requested):
Great functions for atomic database transactions :) 

Overall looks great, just wondering whether or not we should add tests specifically for FIrebaseEventRepository, I think it's possible because we have some examples already in FIrebaseEventRepositoryUnitTest and FirebaseUserRepositoryUnitTest - what do you think?

Other than that, just some nits about error checking, and don't forget to remove the commented code in the end :)

app/src/main/java/com/github/se/eventradar/model/repository/event/FirebaseEventRepository.kt
Viccha18: oh right, there was initially a reason. When I first coded it, incrementPurshases and decrementPurshases where using the same logic so this is why I separated. But afterwards I realized I needed a check on the ticket capacity. So now the addSubPurchases doesn't make sense anymore. I will remove it. Thanks for pointing this out !

app/src/main/java/com/github/se/eventradar/model/repository/event/MockEventRepository.kt
Viccha18: Good point, I should ask the person who coded getEvent what they had in mind before making any changes though

Viccha18:
I think addind some tests wouldn't hurt. I was in the process to do some 👍

app/src/main/java/com/github/se/eventradar/model/repository/event/MockEventRepository.kt
```diff
@@ -81,4 +82,64 @@ class MockEventRepository : IEventRepository {
...
+        Resource.Failure(res.throwable)
+      }
+    }
+  }
+
+  override suspend fun decrementPurchases(eventId: String): Resource<Unit> {
+    return when (val res = getEvent(eventId)) {
+      is Resource.Success -> {
+        val ticket = res.data?.ticket
+        if (ticket != null) {
```
nikhen-s, line 134: nit: maybe include a check here too if purchases is equal to zero, so that it cannot be decremented anymore

app/src/main/java/com/github/se/eventradar/model/repository/user/FirebaseUserRepository.kt
```diff
@@ -182,6 +183,29 @@ class FirebaseUserRepository(db: FirebaseFirestore = Firebase.firestore) : IUser
     }
   }
 
+  override suspend fun addAttendingEvent(userId: String, attendingEventId: String): Resource<Unit> {
```
nikhen-s, line 186: nit: maybe a clearer name would be addAttendeeToEvent (hopefully the refactor feature does it all and it doesn't cause problems, but let me know what you think)

app/src/main/java/com/github/se/eventradar/model/repository/user/FirebaseUserRepository.kt
```diff
@@ -182,6 +183,29 @@ class FirebaseUserRepository(db: FirebaseFirestore = Firebase.firestore) : IUser
...
+    return try {
+      userRef.document(userId).update("eventsAttendeeList", FieldValue.arrayUnion(attendingEventId))
+      Resource.Success(Unit)
+    } catch (e: Exception) {
+      Resource.Failure(e)
+    }
+  }
+
+  override suspend fun removeAttendingEvent(
```
nikhen-s, line 195: nit: likewise for this, maybe removeAttendeeFromEvent, but let me know what you think

nikhen-s (changes requested):
Just some nits! Other than that, LGTM!

app/src/main/java/com/github/se/eventradar/viewmodel/EventDetailsViewModel.kt
```diff
@@ -99,159 +103,77 @@ constructor(
       }
 
       if (!_isUserAttending.value) {
-        // TODO needs to be atomic to avoid concurrency issues
-        registrationUpdateEvent()
-        if (!errorOccurred.value) {
-          registrationUpdateUser()
+        if (registrationProcedure(eventId, currentUserId)) {
+          registrationSuccessful.value = true
```
nikhen-s, line 107: question: is this still needed? it seems in the file, it's only called here

app/src/main/java/com/github/se/eventradar/model/repository/event/MockEventRepository.kt
Viccha18: sure ! ty !

app/src/main/java/com/github/se/eventradar/model/repository/user/FirebaseUserRepository.kt
Viccha18: yeah definetly needs an another name since it's "adding event to user attendance list" xD

app/src/main/java/com/github/se/eventradar/viewmodel/EventDetailsViewModel.kt
Viccha18: yes, it is used by the UI in EventDetails to display the Success Dialog box

nikhen-s (approved):
LGTM!

PR #323, opened 2024-05-26
SCAN FRIEND VM UPDATE

wisdomseeker2001:
There was a race condition due to calling os a suspend function from within the init block.
I have made the necessary changes and there is now no race condition.
The coach approves this method and thinks good practices are used.


app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanFriendQrViewModel.kt
```diff
@@ -121,16 +211,18 @@ constructor(
...
-  fun resetNavigationEvent() {
-    _uiState.value = _uiState.value.copy(action = Action.None)
+  private fun changeAction(action: Action) {
+    _uiState.value = uiState.value.copy(action = action)
+    if (action == Action.NavigateToNextScreen) {
+      navigationActions.navController.navigate(
+          "${Route.PRIVATE_CHAT}/${_uiState.value.decodedResult}") // TODO check correct
+      _uiState.value = _uiState.value.copy(action = Action.None)
+      changeTabState(Tab.MyQR) // TODO add test for this
+    }
```
RaulinN, line 221: Make sure to address these todos beofre merging

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanFriendQrViewModel.kt
```diff
@@ -42,33 +65,100 @@ constructor(
...
+              viewModelScope,
+              SharingStarted.WhileSubscribed(5000),
+              QrCodeScanFriendState(isLoading = true))
+
+  init {
+    viewModelScope.launch { initialUiState.collect { newState -> _uiState.value = newState } }
   }
 
+  //    qrCodeAnalyser.onDecoded = { decodedString ->
+  //      Log.d("QrCodeFriendViewModel", "Decoded QR Code: $decodedString")
```
RaulinN, line 127: Remove this block of 40 lines of commented code if useless

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanFriendQrViewModel.kt
```diff
@@ -121,16 +211,18 @@ constructor(
     return updateResult is Resource.Success
   }
 
-  fun resetNavigationEvent() {
-    _uiState.value = _uiState.value.copy(action = Action.None)
+  private fun changeAction(action: Action) {
```
RaulinN, line 214: (small thing) I cannot highlight the lines above, but the variable "uiLength" should probably be either a constant at the start of the file or somewhere else if you use this threshold elsewhere in the app

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanFriendQrViewModel.kt
```diff
@@ -42,33 +65,100 @@ constructor(
...
+                callbackFlow {
+                      qrCodeAnalyser.onDecoded = { decodedString ->
+                        trySend(decodedString ?: "Failed to decode QR Code")
+                        close()
+                      }
+                      awaitClose { qrCodeAnalyser.onDecoded = null }
+                    }
+                    .first()
+
+            if (decodedResult == "Failed to decode QR Code") {
```
RaulinN, line 106: I would try to change this part. It feels strange to introduce something that is not type safe in kotlin. Even just making it so that the default value is `null` would be better. If you *really* want to keep the string, you might want to create a constant for that

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanFriendQrViewModel.kt
```diff
@@ -121,16 +211,18 @@ constructor(
...
   }
 
-  fun resetNavigationEvent() {
-    _uiState.value = _uiState.value.copy(action = Action.None)
+  private fun changeAction(action: Action) {
+    _uiState.value = uiState.value.copy(action = action)
+    if (action == Action.NavigateToNextScreen) {
+      navigationActions.navController.navigate(
+          "${Route.PRIVATE_CHAT}/${_uiState.value.decodedResult}") // TODO check correct
+      _uiState.value = _uiState.value.copy(action = Action.None)
```
RaulinN, line 219: Ok this is something you use quite often in your code (whole project, everyone), this is not thread safe. Here's an example for a better version:

```kotlin
class CounterModel {
    private val _counter = MutableStateFlow(0) // private mutable state flow
    val counter = _counter.asStateFlow() // publicly exposed as read-only state flow

    fun inc() {
        _counter.update { count -> count + 1 } // atomic, safe for concurrent use
    }
}
```

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanFriendQrViewModel.kt
```diff
@@ -42,33 +65,100 @@ constructor(
...
   }
 
-  private var myUID = ""
+  data class QrCodeScanFriendState(
+      val decodedResult: String = "",
+      val action: Action = Action.None,
+      val tabState: Tab = Tab.MyQR,
+      val username: String = "",
+      val qrCodeLink: String = "",
+      val isLoading: Boolean = true, // Indicates loading state
```
RaulinN, line 74: 👍

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanFriendQrViewModel.kt
wisdomseeker2001: yes still waiting on Achille to confirm if he has got rid of adding extra information to encoding... will actually be deleted soon

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanFriendQrViewModel.kt
wisdomseeker2001: okay, I changed it :)

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanFriendQrViewModel.kt
RaulinN: FOLLOWUP due to a question regarding this:

---

Nah the atomic thing is more general. It does not concern the navigation, but **view models** in general. 

What we often find in EventRadar is something like `_state.value = state.value.copy(modifiedField = modifiedValue)`. What we need to realise, is that we make a COPY of the current state of `state` and THEN assign this value back to `_state.value`. In other words, it is the same thing as if you were doing

```kotlin
// Note : I am assuming that state.value is mutable in the second line just to make the racing condition point clearer, but the problem is exactly the same if state.value is immutable (because we make a *copy* and *then* assign it back)
val oldValue = state.value
oldValue.modifiedField = modifiedValue
_state.value = oldValue
```

What happens when you have the main thread and a coroutine executing this code? Most of the time, everyhting happens correctly. However, they might sometimes negate each other, e.g.

```kotlin
// thread 1 => makes a copy of `state`
val valueT1 = state.value
valueT1.modifiedField = modifiedValue
_state.value = valueT1 

// thread 2 => makes a copy of `state` w/ the modified state from T1
val valueT2 = state.value // valueT1 in this case
valueT2.modifiedField = modifiedValue
_state.value = valueT2

// Here, _state.value ends up with value T2
```

Then, here's another example where the execution differs a bit

```kotlin
// thread 1 => makes a copy of `state`
val valueT1 = state.value
valueT1.modifiedField = modifiedValue

// thread 2 => makes a copy of `state` WITHOUT T1 changes
val valueT2 = state.value
valueT2.modifiedField = modifiedValue
_state.value = valueT2

// thread1 => finishes the execution (maybe the copy was slow, or the field modified was larger)
_state.value = valueT1

// Here, _state.value ends up with value T1. T
```

Same code, two executions, two different results => that's a problem! Now luckily, you can *very simply* resolve this issue by using `.update` on a mutable state flow, e.g.

```kotlin
_state.update{ it.copy(modifiedField = modifiedValue) }
```

If we take the analogy from before, here's what it would do:

```kotlin
synchronize {
    val oldValue = _state.value
    oldValue.modifiedField = modifiedValue
    _state.value = oldValue
}
```

In other words, `.update` performs the change atomically. Meaning that you won't have any race condition.

Lastly, I want to say that this **is** something that could very well happen in your code since you are sometimes using `LaunchedEffect`. Remember that launch effect *listens* for changes, and then **starts a coroutine** when a change is detected on a particular key

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanFriendQrViewModel.kt
achillet3, line 1: The changes look good just a few nits. Nicolas's comments should definitely be taken into account and the update of the UI state should be done with the update function. You could also eventually refactor the adding of both users to their respective friends lists into a function taking the user and friend id to be added.

achillet3 (approved):
Code looks good to me and behaves well on phone. Well done! There is simply the comment from Nicolas to be taken into account.

app/src/main/java/com/github/se/eventradar/viewmodel/qrCode/ScanFriendQrViewModel.kt
wisdomseeker2001: thanks, there actually a reason I did not do so (refactor it this way)... I forgot but I am pretty sure it was because it simplified some code and made it more efficient to batch it in one go. Other chnages made!

PR #333, opened 2024-05-27
Bug fix - nav bar overflow text

christinehahaa:
Limited nav bar text to 1 line and added elipsis when text overflows.


christinehahaa:
Accidentally didn't branch off most updated main, causing some build issues. Replaced with new PR #334

PR #348, opened 2024-05-29
Ensuring more atomicity 

wisdomseeker2001:
Chnaged to state.update in all VMs!

### DIFF
--- a/app/src/main/java/com/github/se/eventradar/viewmodel/AuthenticationViewModel.kt
+++ b/app/src/main/java/com/github/se/eventradar/viewmodel/AuthenticationViewModel.kt
@@ -19,6 +19,7 @@ import java.util.Locale
 import javax.inject.Inject
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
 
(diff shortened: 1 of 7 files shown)


Viccha18 (approved):
LGTM, I guess this is what happen when a team doesn't know about best practices and we kept copying each other ways of coding things. Thanks for the tedious work ! 👍

app/src/main/java/com/github/se/eventradar/viewmodel/AuthenticationViewModel.kt
```diff
@@ -132,7 +133,9 @@ class LoginViewModel @Inject constructor(private val userRepository: IUserReposi
   ) {
     when (val result = userRepository.addUser(userValues, userId)) {
       is Resource.Success -> {
-        state.value = state.value.copy(isSignUpCompleted = true, isSignUpSuccessful = true)
+        state.update { currentState ->
```
jgeaso1266, line 136: Is there a reason why you use the `currentState` here and `it` elsewhere?

app/src/main/java/com/github/se/eventradar/viewmodel/MessagesViewModel.kt
```diff
@@ -58,7 +58,8 @@ constructor(
               }
 
           // Update the UI state with the sorted and filtered message list
-          _uiState.value = _uiState.value.copy(messageList = sortedMessageList)
+
```
jgeaso1266, line 61: nit: space should be deleted

jgeaso1266 (commented):
Nice, but the use of `currentState` vs `it` seems random. I suggest they be the same for consistency. Also, this change should also be made in the tests.

PR #352, opened 2024-05-29
Added display message when no filtered results + test updates

christinehahaa:
Main changes:
- "No events match the filter applied" text displayed in map and list view on home page + on my hosted events page when filtering.
- Added 2 UI tests, 1 to Hosting and 1 to Home to check the message displays correctly.


app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
```diff
@@ -288,6 +288,11 @@ constructor(
     // Reset search and filter when tab is changed
     state.value =
```
wisdomseeker2001, line 289: you should always be using state.update {} to ensure atmocity and avoid race conditions, I made a PR which changed this for all VMs so you do not need to make this change

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
```diff
@@ -288,6 +288,11 @@ constructor(
...
         state.value.copy(
+            eventList =
```
wisdomseeker2001, line 291: could u please explain wjyu within the state.value.copy, you have that...

upcomingEventList and upcomingEventList  = to a state.value.copy...?

I tried going on ur branch to make it more optimal so i wanted to aks if there is some reason for this that I am not getting

wisdomseeker2001 (approved):

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
christinehahaa: Great thanks, noted.

app/src/main/java/com/github/se/eventradar/viewmodel/EventsOverviewViewModel.kt
christinehahaa: Good point, I have updated to use the currentState.update value instead of copy.

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -70,7 +70,17 @@ fun HostingScreen(
   GetUserLocation(context, viewModel::onUserLocationChanged)
 
   ConstraintLayout(modifier = Modifier.fillMaxSize().testTag("hostingScreen")) {
-    val (logo, title, divider, searchfilter, filter, eventList, eventMap, bottomNav, buttons) =
+    val (
```
wisdomseeker2001, line 73: not sure why it has been formatted like this in comparison to before hahah but not counted as new lines in SOnarCLoud so all good

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
```diff
@@ -102,6 +112,22 @@ fun HostingScreen(
         placeholderStringResource = R.string.hosting_search_placeholder)
 
     val events = getEventsBasedOffUiState(uiState)
+    var mapTopConstraint = searchfilter.bottom
```
wisdomseeker2001, line 115: this indeed the proper use of constraind layout childs - seems all good to me, only improvement I can suggets but do not insist on is doing as done for HomeScreen etc where the items displayed are actually composable functions in ReusableComponents. so for excample you would have composable function in Reusabele components called something along the lines of TextFilterButNoMatch(modifier: Modifier) { } and then you would just call it at line 119 - makes eveything more clean and organised. should take u 8 mins, no more

wisdomseeker2001:
Overall I am happy with the changes, I will approve but would advise you to reformat the code based on my above suggestion

wisdomseeker2001 (approved):

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
christinehahaa: Not sure either, kfmt did it

app/src/main/java/com/github/se/eventradar/ui/hosting/Hosting.kt
christinehahaa: I made these changes and it only neatens it by two lines, since the modifier has to be passed in still. Therefore, I have elected not to make these changes.

PR #357, opened 2024-05-30
Bug fix - FABs on "My Hostings" page

christinehahaa:
Fixed bug related to "create event" and "map/list" button floating at top of screen when search is active.


wisdomseeker2001 (approved):
Very small PR, just some layout linking among different children of the Constraint Layout: LGTM

PR #359, opened 2024-05-30
UX/UI Refinement

christinehahaa:
Some refinement changes to the UX / UI, namely:
- Added "No" button to event cancel dialog box
- ... tbc

NOTE: this is a draft PR, cannot be finished until everyone's done their PRs and merged into main


app/src/main/java/com/github/se/eventradar/ui/qrCode/QrCodeTicketUi.kt
```diff
@@ -140,23 +139,26 @@ fun QrCodeTicketUi(
...
-              MaterialTheme.colorScheme.onSurface,
-          )
-
       val (lazyEventDetails) = createRefs()
 
-      // TODO uncomment when image are implemented
-      // val imagePainter: Painter = rememberImagePainter(eventUiState.eventPhoto)
-      val imagePainter: Painter = rememberImagePainter(R.drawable.placeholderbig)
+      val imagePainter =
+          if (qrScanUiState.value.eventUiState.eventPhoto == "") {
```
wisdomseeker2001, line 157: this is causing a bug, just tried when running preview, please try and fix. I would suggest to do "
val imagePainter =
           if (qrScanUiState.value.eventUiState.eventPhoto == "") {
rememberImagePainter(eventUiState.eventPhoto) else {
rememberAsyncImagePainter(qrScanUiState.value.eventUiState.eventPhoto)} and see if that works

app/src/main/java/com/github/se/eventradar/ui/qrCode/QrCodeTicketUi.kt
```diff
@@ -221,6 +220,17 @@ fun QrCodeTicketUi(
             item { Spacer(modifier = Modifier.height(8.dp)) }
           }
     } else {
+      EventTitle(
```
wisdomseeker2001, line 223: personally do not agree wiht the UI choice here, does not make sense in my personal opinion to have the event title here, and I think makes the UI less nice. But these are simply UI preferenecs so if u wanna keep it like that, so be it :)

wisdomseeker2001 (changes requested):
Overall thank you for chnaging distance to location makes a lot of sense and all other chnages, besides my comment LGTM

wisdomseeker2001 (approved):
All good now,

PR #274, opened 2024-05-15
Add intents rule and hilt injection rule

jgeaso1266:

PR #363, opened 2024-06-01
Instantiation tests for assisted injected viewModels

Viccha18:
# Objective

Test the instantiation of an assisted-injected viewModel using the create function from the viewModels' companion object.

# Note
Reintroducing + extending the code from https://github.com/swent-sp-2024-event-radar/event-radar-app/pull/275 that what merged (because of similarities) into https://github.com/swent-sp-2024-event-radar/event-radar-app/pull/274 which has been discarded because it contained redundancies with the rest of the code.


christinehahaa (approved):
A small PR with some extra testing, LGTM!

PR #365, opened 2024-06-02
Edit profile vm+UI

achillet3:
# Objective
Implementation of the Edit Profile Screen. It has been merged into the profile screen and activated upon clicking the edit icon. 

# Implementation
- Upon clicking the edit icon the UI displays outlined text fields instead of texts
- A save button is present at the bottom of the screen to update the user info in the database

# Additional notes
- Left to implement: Log out button and delete account button

### DIFF
--- a/app/src/androidTest/java/com/github/se/eventradar/viewProfile/ProfileTest.kt
+++ b/app/src/androidTest/java/com/github/se/eventradar/viewProfile/ProfileTest.kt
@@ -114,42 +114,6 @@ class ProfileTest : TestCase(kaspressoBuilder = Kaspresso.Builder.withComposeSup
   }
 
   @Test
-  fun screenDisplaysAllElementsCorrectlyWhenPrivateInEditMode() = run {
-    composeTestRule.setContent {
-      ProfileUi(
-          isPublicView = false,
-          viewModel = mockProfileViewModel,
-          navigationActions = mockNavActions)
-    }
-
-    ComposeScreen.onComposeScreen<ProfileScreen>(composeTestRule) {
-      editButton {
-        // arrange: verify the pre-conditions
-        assertIsDisplayed()
-
-        // act: click the edit button
-        performClick()
-      }
-      goBackButton { assertIsDisplayed() }
-      editProfile { assertIsDisplayed() }
-      profilePic {
-        assertIsDisplayed()
-        assertHasClickAction()
-      }
-      firstNameTextField { assertIsDisplayed() }
-      lastNameTextField { assertIsDisplayed() }
-      usernameTextField { assertIsDisplayed() }
-      bioTextField { assertIsDisplayed() }
-      phoneNumberTextField { assertIsDisplayed() }
-      birthDateTextField { assertIsDisplayed() }
-      saveButton {
-        assertIsDisplayed()
-        assertHasClickAction()
-      }
-    }
-  }
-
-  @Test
   fun goBackButtonTriggersBackNavigation() = run {
     composeTestRule.setContent {
       ProfileUi(
--- a/app/src/androidTest/java/com/github/se/eventradar/viewProfile/ProfileTest.kt
+++ b/app/src/androidTest/java/com/github/se/eventradar/viewProfile/ProfileTest.kt
@@ -1,8 +1,11 @@
 package com.github.se.eventradar.viewProfile
 
+import android.content.Intent
+import android.net.Uri
 import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.test.espresso.intent.Intents
+import androidx.test.espresso.intent.matcher.IntentMatchers
 import androidx.test.ext.junit.runners.AndroidJUnit4
-import com.github.se.eventradar.model.User
 import com.github.se.eventradar.screens.ProfileScreen
 import com.github.se.eventradar.ui.navigation.NavigationActions
 import com.github.se.eventradar.ui.viewProfile.ProfileUi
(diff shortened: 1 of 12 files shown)


wisdomseeker2001 (approved):
Went thought all your code, did not see anything out of the ordinary, you respected all the good practices that I know of, works on emulator. Really goood job that I did not have any comments for such big code! You killed it

app/src/main/java/com/github/se/eventradar/ui/viewProfile/ProfileUi.kt
```diff
@@ -102,6 +136,20 @@ fun ProfileUi(
...
+            Icon(
+                painter = painterResource(id = R.drawable.edit),
+                contentDescription = "Edit Button",
+                modifier = Modifier.size(32.dp),
+                tint = MaterialTheme.colorScheme.onPrimaryContainer,
+            )
+          }
         }
       },
       content = {
```
nikhen-s, line 155: Currently, the keyboard covers the textfields that a user is currently editing, I think maybe it could be fixed by adding a scroll state in the modifier of the column :)
` .verticalScroll(rememberScrollState())`

app/src/main/java/com/github/se/eventradar/ui/viewProfile/ProfileUi.kt
```diff
@@ -113,68 +161,183 @@ fun ProfileUi(
...
+                                        unfocusedLabelColor = MaterialTheme.colorScheme.primary,
+                                    ))
+                      }
+                }
+
+                Button(
+                    onClick = {
+                      if (viewModel.validateFields()) {
+                        viewModel.updateUserInfo()
+                        isEditMode.value = false
```
nikhen-s, line 270: Currently, when the user clicks the goBack button (which I saw sets the isEditMode.value to false), it somehow triggers this? and the user information still gets saved, maybe a fix could be adding a state in the VM for isCancelSelected (which is updated to true when user clicks go back button). I'm not sure if this would fix it all though but hope it helps!

app/src/main/java/com/github/se/eventradar/ui/viewProfile/ProfileUi.kt
```diff
@@ -113,68 +161,183 @@ fun ProfileUi(
...
+                      color = MaterialTheme.colorScheme.onSurface,
+                      fontSize = 16.sp,
+                      style = MaterialTheme.typography.bodyMedium)
+                  if (!isPublicView) {
+                    Row(
+                        modifier = Modifier.padding(top = 16.dp).testTag("phoneNumberBirthDateRow"),
+                        horizontalArrangement = Arrangement.SpaceBetween) {
+                          Column(modifier = Modifier.weight(1f).testTag("phoneNumberColumn")) {
+                            Text(
+                                text = "Phone Number",
```
nikhen-s, line 303: nit: maybe a nice feature to implement would be to include the country code but it might take too long, so not a priority

app/src/main/java/com/github/se/eventradar/ui/viewProfile/ProfileUi.kt
```diff
@@ -113,68 +161,183 @@ fun ProfileUi(
...
+                                text = uiState.phoneNumber,
+                                modifier =
+                                    Modifier.padding(top = 4.dp).testTag("phoneNumberInfoText"),
+                                color = MaterialTheme.colorScheme.onSurface,
+                                fontSize = 16.sp,
+                                style = MaterialTheme.typography.bodyMedium)
+                          }
+                          Column(modifier = Modifier.weight(1f).testTag("birthDateColumn")) {
+                            Text(
+                                text = "Birth Date",
```
nikhen-s, line 319: suggestion: could add a placeholder for the birth date format here :)

nikhen-s (changes requested):
The UI looks great :) Overall looks great, but noticed a few small bugs:

- When editing a textfield, the texfield gets covered by the keyboard, I think the issue might be that the Column is not scrollable, included it as a comment.
- Upon clicking on the GoBack button in Edit Profile, the Profile actually gets edited if I changed a textfield, (whereas I think the intended behavior is to cancel the edits and only save upon clicking the save button).
- Currently, all the user birth dates are in the form dd/MM/yyyy which means that if a user edits their profile it will always be red (which might confuse the user), maybe we could change all the dates we have right now to the form dd.MM.yyyy in firebase, and also include a placeHolder in the date textfield.

app/src/androidTest/java/com/github/se/eventradar/viewProfile/ProfileTest.kt
```diff
@@ -30,22 +35,40 @@ class ProfileTest : TestCase(kaspressoBuilder = Kaspresso.Builder.withComposeSup
...
+
+    mockProfileViewModel = ProfileViewModel(mockUserRepository, "1")
+
+    mockProfileViewModel.getProfileDetails()
   }
 
+  @After fun testTeardown() = runTest { unmockkAll() }
+
   @Test
   fun screenDisplaysAllElementsCorrectlyWhenPublic() = run {
```
christinehahaa, line 73: Nit: The name of this is a little confusing, what do you mean by "WhenPublic"?

app/src/androidTest/java/com/github/se/eventradar/viewProfile/ProfileTest.kt
```diff
@@ -63,47 +86,30 @@ class ProfileTest : TestCase(kaspressoBuilder = Kaspresso.Builder.withComposeSup
...
-      phoneNumberInfoText { assertDoesNotExist() }
-      birthDateColumn { assertDoesNotExist() }
-      birthDateLabelText { assertDoesNotExist() }
-      birthDateInfoText { assertDoesNotExist() }
     }
   }
 
   @Test
-  fun screenDisplaysAllElementsCorrectlyWhenPrivate() = run {
+  fun screenDisplaysAllElementsCorrectlyWhenPrivateNotInEditMode() = run {
```
christinehahaa, line 95: Nit: same as this name, what does "WhenPrivate" mean?

app/src/main/java/com/github/se/eventradar/ui/viewProfile/ProfileUi.kt
```diff
@@ -113,68 +161,183 @@ fun ProfileUi(
...
+                      .clip(RoundedCornerShape(10.dp))
+                      .clickable(enabled = isEditMode.value) {
+                        imagePickerLauncher.launch("image/*")
+                      })
+              if (isEditMode.value) {
+                // Display text fields for editing the user's first and last name on the same row
+                Row(
+                    horizontalArrangement = Arrangement.SpaceBetween,
+                    modifier = Modifier.fillMaxWidth().testTag("nameRow")) {
+                      OutlinedTextField(
```
christinehahaa, line 180: Restrict to 1 line with `singleLine = true,`

app/src/main/java/com/github/se/eventradar/ui/viewProfile/ProfileUi.kt
```diff
@@ -113,68 +161,183 @@ fun ProfileUi(
...
+                          label = { Text("First Name") },
+                          isError = uiState.firstNameIsError,
+                          colors =
+                              OutlinedTextFieldDefaults.colors()
+                                  .copy(
+                                      focusedPrefixColor = MaterialTheme.colorScheme.primary,
+                                      unfocusedLabelColor = MaterialTheme.colorScheme.primary,
+                                  ))
+                      Spacer(modifier = Modifier.width(10.dp).testTag("nameSpacer"))
+                      OutlinedTextField(
```
christinehahaa, line 193: Restrict to 1 line with `singleLine = true,`

app/src/main/java/com/github/se/eventradar/ui/viewProfile/ProfileUi.kt
```diff
@@ -113,68 +161,183 @@ fun ProfileUi(
...
+                          label = { Text("Last Name") },
+                          isError = uiState.lastNameIsError,
+                          colors =
+                              OutlinedTextFieldDefaults.colors()
+                                  .copy(
+                                      focusedPrefixColor = MaterialTheme.colorScheme.primary,
+                                      unfocusedLabelColor = MaterialTheme.colorScheme.primary,
+                                  ))
                     }
+                OutlinedTextField(
```
christinehahaa, line 206: Restrict to 1 line with `singleLine = true,`

app/src/main/java/com/github/se/eventradar/ui/viewProfile/ProfileUi.kt
```diff
@@ -113,68 +161,183 @@ fun ProfileUi(
...
+                            isError = uiState.phoneNumberIsError,
+                            colors =
+                                OutlinedTextFieldDefaults.colors()
+                                    .copy(
+                                        focusedPrefixColor = MaterialTheme.colorScheme.primary,
+                                        unfocusedLabelColor = MaterialTheme.colorScheme.primary,
+                                    ))
+                        Spacer(
+                            modifier = Modifier.width(10.dp).testTag("phoneNumberBirthDateSpacer"))
+                        OutlinedTextField(
```
christinehahaa, line 251: Restrict to 1 line with `singleLine = true,`

app/src/main/java/com/github/se/eventradar/ui/viewProfile/ProfileUi.kt
```diff
@@ -113,68 +161,183 @@ fun ProfileUi(
...
+                      colors =
+                          OutlinedTextFieldDefaults.colors()
+                              .copy(
+                                  focusedPrefixColor = MaterialTheme.colorScheme.primary,
+                                  unfocusedLabelColor = MaterialTheme.colorScheme.primary,
+                              ))
+                  Row(
+                      horizontalArrangement = Arrangement.SpaceBetween,
+                      modifier = Modifier.padding(top = 10.dp).testTag("phoneNumberBirthDateRow")) {
+                        OutlinedTextField(
```
christinehahaa, line 237: Restrict to 1 line with `singleLine = true,`

app/src/main/java/com/github/se/eventradar/ui/viewProfile/ProfileUi.kt
```diff
@@ -113,68 +161,183 @@ fun ProfileUi(
...
+                            colors =
+                                OutlinedTextFieldDefaults.colors()
+                                    .copy(
+                                        focusedPrefixColor = MaterialTheme.colorScheme.primary,
+                                        unfocusedLabelColor = MaterialTheme.colorScheme.primary,
+                                    ))
+                      }
+                }
+
+                Button(
```
christinehahaa, line 266: Currently when clicking the save button, it stays on the edit profile screen. Upon clicking "save", the screen should navigate back to the profile screen.

app/src/main/java/com/github/se/eventradar/ui/viewProfile/ProfileUi.kt
christinehahaa: Perhaps not a placeholder, since the existing DOB text is displayed, but add an example below the text input, same as what is done for username having the text "Your username must be unique". Text can be "e.g. 10.12.1997"

app/src/main/java/com/github/se/eventradar/ui/viewProfile/ProfileUi.kt
```diff
@@ -113,68 +161,183 @@ fun ProfileUi(
...
+                                    ))
+                        Spacer(
+                            modifier = Modifier.width(10.dp).testTag("phoneNumberBirthDateSpacer"))
+                        OutlinedTextField(
+                            value = uiState.birthDate,
+                            onValueChange = viewModel::onBirthDateChanged,
+                            modifier = Modifier.weight(1f).testTag("birthDateTextField"),
+                            label = { Text("Birth Date") },
+                            isError = uiState.birthDateIsError,
+                            colors =
```
christinehahaa, line 257: Agree with Nik regarding DOB format, Vicky changed it to use "." not "/". This will need to be changed in the firebase to ensure validation is correct and no crashes when pulling data. 

Also keyboard should be numeric too for DOB.

christinehahaa (changes requested):
Nik has done a thorough review. Alongside his bugs, I tested in app and found some additional ones, which I have commented inline. 

Also, I see the log out function is still to be implemented. This should definitely be added before this PR is merged. The M4 deliverables require the app to be complete end-to-end which I assume means they should be able to end a user flow (i.e. by logging out). Hopefully it's not too difficult to implement?

app/src/main/java/com/github/se/eventradar/ui/viewProfile/ProfileUi.kt
nikhen-s: Good idea! Agree, that would be more user-friendly

app/src/main/java/com/github/se/eventradar/ui/viewProfile/ProfileUi.kt
achillet3: True! Thanks for the catch

app/src/main/java/com/github/se/eventradar/ui/viewProfile/ProfileUi.kt
achillet3: I implemented the goBackButton to save the changes only in the UI state if the user presses the back function. If he decides to finish editing his profile, he can continue where he left off and if he leaves the changes will be lost. I added a Toast message saying the User Info was not saved. I hope this helps clear up the behavior.

app/src/androidTest/java/com/github/se/eventradar/viewProfile/ProfileTest.kt
achillet3: As the description explains, I have merged both the Profile page and the Friends' View Profile, to limit code redundancy. In the test and in the code this is portrayed by the isPublicView boolean that indicates if another user is seeing my "Public" profile or if I am seeing my "Private" Profile

app/src/main/java/com/github/se/eventradar/ui/viewProfile/ProfileUi.kt
achillet3: That's surprising on my end it goes back to the non-edit mode profile screen, which is the profile screen.

app/src/androidTest/java/com/github/se/eventradar/viewProfile/ProfileTest.kt
christinehahaa: Gotcha, thanks for the explanation.

app/src/main/java/com/github/se/eventradar/ui/viewProfile/ProfileUi.kt
christinehahaa: It seems to work fine now!

app/src/main/java/com/github/se/eventradar/ui/viewProfile/ProfileUi.kt
christinehahaa: This looks great. Only thing is the unsaved changes get removed when I switch tabs - is there a way to keep the unsaved changes across the entire app lifecycle, unless the user logs out?

nikhen-s (approved):
LGTM

app/src/main/java/com/github/se/eventradar/ui/viewProfile/ProfileUi.kt
christinehahaa: Noted that this is too difficult to change in the limited time we have. The behaviour will have to remain like this.

christinehahaa (approved):
Thanks for making all the changes. Great work! LGTM.

